PR #23, opened 2015-11-14
Change Json to Erlang Media Type

yohaneskuma:
The dependency towards jsx for encoding and decoding is replaced with Erlang Media Type which uses Maps as data structure.
All the side-effects are cleaned.


tsloughter:
Thanks! There are some indentation issues, like https://github.com/ieug/rebar3_hex/blob/using_erlang_media_type/src/rebar3_hex_user.erl#L169 , could you clean those up and squash the commits?

Only issue I see here is the use of maps that means only 17+ can use the plugin. This is probably fine and actually we may already have only supported R16+ in the first place with this plugin, if I recall it has always failed on R15. But I need to make sure this is the case and is ok.

yohaneskuma:
I have squashed the indentation issues. I hope it is clean now.

PR #28, opened 2015-12-26
Add functionality for "rebar3 hex info"

yohaneskuma:
The current implementation of rebar3_hex has no functionality for command 'rebar3 hex info' if the package is not specified.
In this commit, we have added functionality for this command, which will list out:
    1. All deps at level 0
    2. for those deps at level 0 which are package deps (are found in Hex registry), it will continue   listing out their deps at level > 0

This will be helpful, since the current implementation of 'rebar3 deps' only list out the deps at level 0.

### DIFF
--- a/src/rebar3_hex_info.erl
+++ b/src/rebar3_hex_info.erl
@@ -10,6 +10,8 @@
 -define(DEPS, []).
 
 -define(ENDPOINT, "packages").
+-define(RELEASE, "releases").
+-define(REGISTRY_FILE, "registry").
 
 %% ===================================================================
 %% Public API
@@ -35,7 +37,7 @@ init(State) ->
 do(State) ->
     case rebar_state:command_args(State) of
         [] ->
-            general();
+            general(State);
         [Package] ->
             package(Package);
         [Package, Version] ->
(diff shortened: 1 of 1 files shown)


tsloughter:
@yohaneskuma hey, sorry I'm late on this I somehow missed the PR. It looks like this fetches packages here https://github.com/hexpm/rebar3_hex/pull/28/files#diff-07e7d677eea863060d701c58e8645254R79 and i don't understand why. Isn't all the necessary information available in the local registry?

yohaneskuma:
Yes we could use the local registry. I used it in that way (via fetching) in order to have the same pattern as of the other services of 'rebar3 hex info' command. As you can see here https://github.com/hexpm/rebar3_hex/pull/28/files#diff-07e7d677eea863060d701c58e8645254L53 and https://github.com/hexpm/rebar3_hex/pull/28/files#diff-07e7d677eea863060d701c58e8645254L82 , rebar3_hex_info fetches directly from hex instead of the local registry. Shall we use the local registry for 'rebar3 hex info' , while communicating the hex registry for 'rebar3 hex info PACKAGE [VERSION]' ?

tsloughter:
Oh, yes, we should definitely use the local registry here. Maybe I copied how mix's hex plugin does it without thinking. It would be much better to not fetch.

yohaneskuma:
@tsloughter  I have squashed a new version as of your comment. It will now first use local registry, and if it is missing it will contact hex server. With this pull request, we can have the following functionality:

Given a sample rebar.config 
`{erl_opts, [debug_info]}.
{deps, [cowboy] or [{cowboy, "1.0.4"}]}.
{plugins, [rebar3_hex]}.
`
we can get the following output from performing 'rebar3 hex info':
`cowboy 1.0.4`
`cowlib ~> 1.0.0`
`ranch ~> 1.0
`

src/rebar3_hex_info.erl
```diff
@@ -10,6 +10,8 @@
 -define(DEPS, []).
 
 -define(ENDPOINT, "packages").
+-define(RELEASE, "releases").
+-define(REGISTRY_FILE_LOCATION, "/home/yohanes/.cache/rebar3/hex/default/registry").
```
tsloughter: Need to use `rebar_dir` functions not hardcoded to your dir :)

yohaneskuma:
Ohh you are right. I forgot to change it. I will improve it. Thanks

yohaneskuma:
@tsloughter I have now used `init:get_argument(home)` to get home dir of the user.

tsloughter:
Sorry, I should have been more clear, you can't do that either. You need to do:

``` erlang
{ok, RegistryDir} = rebar_packages:registry_dir(State)
```

yohaneskuma:
I am kinda confused here. Do you think it is good idea to use rebar_packages from rebar3 in rebar3_hex project? aren't rebar3 and rebar3_hex completely separated?

tsloughter:
Oh, no, they aren't. rebar3_hex is a plugin to rebar3, I could move the call to be rebar_api:registry_dir(State) but needs to be used either way.

yohaneskuma:
@tsloughter Thanks, I have used  now `rebar_packages:registry_dir(State)` as you suggested

tsloughter:
Great, looking good. One last nitpick :):

``` erlang
-define(REGISTRY_FILE, "registry").
...
case ets:file2tab(filename:join(RegistryDir, ?REGISTRY_FILE)) of
```

yohaneskuma:
@tsloughter it now uses filename:join. thanks

PR #35, opened 2016-03-09
Fix for hex docs

knusbaum:
rebar3 hex docs blows up when trying to drop "doc" from the short path name.

### DIFF
--- a/src/rebar3_hex_docs.erl
+++ b/src/rebar3_hex_docs.erl
@@ -86,7 +86,7 @@ delete(Name, Version) ->
     end.
 
 file_list(Files) ->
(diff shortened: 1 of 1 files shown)


mremond:
Thanks @knusbaum ! I had the same issue and it helped be able to submit docs update to hex.pm.
@tsloughter It would be super helpful if you could release a new version with that patch. That would make sending docs to hex.pm work out of the box.

Thanks !

tsloughter:
Published, `1.20.0` let me know that it works for you.

knusbaum:
@tsloughter Works for me.

mremond:
👍 I confirm it works as well.

PR #40, opened 2016-04-12
Add include_files and exclude_files options

waisbrot:
Based on @benoitc 's suggestion in #26. If erlcloud/erlcloud#270 is merged to supply a `mix.exs` file then `include_files` would be nice to have.

`include_files` are included in addition to what's specified in `files`
(or the default set of files). `exclude_files` are expanded and removed from the
set of files to publish after `include_files` have been included.


waisbrot:
I saw you move @tsloughter :-)  Do you have any comments on this one? I could add some eunit tests or try some alternative implementations if that would help.

tsloughter:
Oh right, let me look at the code again to be sure and then i'll merge.

tsloughter:
Looks good. Sadly I can't expect tests for anything since I never added tests for anything else yet :). I'll need to do that some day so that testing isn't just expected by the end user, hehe.

PR #47, opened 2016-05-12
Read `build_tools` from .app data and show more metadata

### LINKED ISSUE #36: Don't ignore build_tools
If `build_tools` is in the `.app.src` include it (along with `rebar3`) in the package metadata.

waisbrot:
Read from the `build_tools` key, if it's present, or default to
`[<<"rebar3">>]` if it's not.

Print the description, maintainers, licenses, links, and build tools
when prompting the user about publishing the package.

Closes #36


waisbrot:
I added the printing because I recently published a package (https://hex.pm/packages/dogstatsde) and it doesn't have any build tools listed. I can see, though, that `<<"rebar3">>` is being sent, so I'm not sure if that's a Hex problem.

tsloughter:
Great, thanks!

Issue #41, opened 2016-04-16
ignore known temp files


Licenser:
things like:

```
*~
.#*
...
```

that are know to be markers for temp files, it would be nice if rebar3_hex would ignore them by default

starbelly:
@tsloughter I think this can be closed

Issue #55, opened 2016-08-03
Don't publish compiled binaries


Licenser:
when a c_src directory exists rebar_hex will publish .d, .o, .so objects along with the source, that can lead to very bad results (aka loosing platform independence). Making this mistake is much too easy.

PR #70, opened 2017-08-08
Rework exclude logic for hex packages

Licenser:
This improves excluding files from hex uploads. In total there are three changes here

1) Explicit includes overwrite excludes, this is so regexp excludes don't need to be overly complicated and users can still say "I really want this one specific file".
2) `exclude_regexps` in the app file can now hold regular expressions that exclude files based on a pattern (for example `"\\.d$"` for c-code that we don't want to always exclude is someone might write d (sure that will happen ...))
3) A list of known bad files that we don't want to upload in hex packages ever, this includes emacs and vim temp files, as well as compiled c-code. (those *can* be overwritten with explicit includes if needed)

### DIFF
--- a/src/rebar3_hex_pkg.erl
+++ b/src/rebar3_hex_pkg.erl
@@ -114,6 +114,23 @@ publish(App, State) ->
             Error
     end.
 
+
+known_exclude_file(Path, ExcludeRe) ->
+    KnownExcludes = [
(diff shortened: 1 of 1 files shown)


Licenser:
This would resolve #41 and #55

Licenser:
ping

PR #74, opened 2017-10-11
Support Unicode strings in app. description and package metadata

g-andrade:
Hi,

I caught this upon trying to publish a package which included a (>255) Unicode codepoint in the application description - it triggered a crash.


g-andrade:
I do have some reservations about how wild it can get if any symbol is allowed; if you think it best, it's feasible to restrict codepoints to specific Unicode classes. On the other hand, one could argue that it's rather be up to the final destination to do so (hex.pm), and hopefully they're already filtering out most irrelevant symbols, emoji and visual abuses on their side.

g-andrade:
[bump] What do you think?

tsloughter:
Thanks! Sorry I forgot to merge this.

tsloughter:
And right, for now I'd prefer to assume hex.pm will do the necessary filtering if it doesn't want certain symbols.

PR #75, opened 2017-10-19
Fix `rebar3 hex info *` commands

gomoripeti:
- hex info commands were failing with authentication error (probably because some recent change on hex.pm server)
- printing the error message was also crashing


tsloughter:
Sorry about not getting back on this. It is now very outdated by the move to `hex_core`.

gomoripeti:
obviously, thanks anyway for the new hex plugin

PR #80, opened 2018-08-11
Allow to publish without prompt. Useful for scripts

tank-bohr:
I'am trying to set up automatic package publishing in CI on tag pushed event. Thus this thing is crucial to making it possible

### DIFF
--- a/src/rebar3_hex_pkg.erl
+++ b/src/rebar3_hex_pkg.erl
@@ -5,7 +5,7 @@
          format_error/1]).
 
 -export([publish/2
-        ,publish/6
+        ,publish/7
         ,validate_app_details/1]).
 
 -include("rebar3_hex.hrl").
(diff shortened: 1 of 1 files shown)


tsloughter:
You can get around it for now with http://blog.erlware.org/automatic-hex-package-publishing-with-travis-ci/

I'll think about this. My thinking was that these extra steps keep it from being done in regular use, viewing the output is important and can catch plenty of mistake -- I do all the time at least :).

tank-bohr:
@tsloughter Thank you. I'll take your advice

tank-bohr:
@tsloughter What will we do about it? I still believe it worth to merge. This option is false by default thus you still can catch the mistakes.
But if you don't think so then we should just close this PR

tsloughter:
This actually isn't possible anymore because the key for publishing is encrypted on disk, requiring a password from the user to decrypt at the time of publishing. This is to prevent issues like when an npm package was stealing credentials.

PR #82, opened 2018-12-14
Return and test friendly error messages on registration failure

starbelly:
This PR introduces a mock hex api web server for "integration" testing the rebar3 hex commands. 
Calls that would require user input from stdin are mecked. 
 The hex api mock has db service (gen_server  with ets) to go along with it to faciliate simulating an already registered user. 

 More PRs will follow building upon this framework to support resolution of existing and future issues.


starbelly:
@tsloughter Oi, this is ready.

PR #85, opened 2018-12-30
Hard deprecrate maintainers field and validate required fields prior to publishing

### LINKED ISSUE #20: Check description is not empty on publish
In order to prevent packages being published, we need to alert users that they need to include a description (see hexpm/hex#145 and hexpm/hex_web#162)

This can be fixed [here](https://github.com/hexpm/rebar3_hex/blob/848b8d3e42063168eeaf12a8fa44762acc162121/src/rebar3_hex_pkg.erl#L106), I assume

starbelly:
- Closes #20, #44 
 - Removes maintainers from metadata


starbelly:
@tsloughter This is ready.

src/rebar3_hex_publish.erl
```diff
@@ -8,14 +8,20 @@
...
+        validate_app_details/1]).
 
 -include("rebar3_hex.hrl").
 
 -define(PROVIDER, publish).
 -define(DEPS, [{default, lock}]).
 
+-define(VALIDATIONS, [  has_semver 
+                      , has_contributors
+                      , has_maintainers
```
wojtekmach, line 21: fyi: maintainers is deprecated too :) It's ignored on the server. See rationale: https://github.com/hexpm/hex/blob/master/CHANGELOG.md#ignoring-maintainers-field

wojtekmach (commented):
Hi @starbelly, this is great. If we can catch errors before even hitting the server that's definitely the way to go.

We have plans to add metadata validations in https://github.com/hexpm/hex_core/issues/22, and the idea is they'd be shared between hex, hex.pm, rebar3 and other tooling. We could start with something really basic and move more validations over time. If you'd like to help with that in hex_core, let us know!

cc @ericmj

src/rebar3_hex_publish.erl
starbelly: Thank you for that catch. I'll check out the issue, but yes I'll be spending a good bit of time on all this over the next few weeks at least, so PRs to hex_core are on the way 👍

starbelly: Resolved via removing those validations, safe to ignore.

tsloughter:
Merging this. But yes, would be great to get this into hex_core instead.

PR #86, opened 2019-01-03
Prompt for secure input on win32 ala hex

starbelly:
- A bug where we don't use ?PVR_ERROR on password mismatch was also fixed which affected all platforms.

### DIFF
--- a/src/rebar3_hex_utils.erl
+++ b/src/rebar3_hex_utils.erl
@@ -53,6 +53,14 @@ format_error({not_valid_repo, RepoName}) ->
 -define(OP_PUTC, 0).
 
 get_password(Msg) ->
+    case os:type() of
+        {win32, nt} -> 
+            get_win32_password(Msg);
+         _ ->
+            get_passwd(Msg)
+    end.
+
+get_passwd(Msg) ->
     case io:setopts([binary, {echo, false}]) of
         ok ->
             PwLine = io:get_line(Msg),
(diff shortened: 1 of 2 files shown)


src/rebar3_hex_utils.erl
```diff
@@ -53,27 +53,54 @@ format_error({not_valid_repo, RepoName}) ->
...
-                    error_logger:tty(true),
-                    Pw
+            case io:setopts([binary, {echo, false}]) of
+                ok ->
+                PwLine = io:get_line(Msg),
+                ok = io:setopts([binary, {echo, true}]),
+                io:format("\n"),
+                [Pw | _] = binary:split(PwLine, <<"\n">>),
+                Pw;
+            _ ->
```
tsloughter: should be indented more?

src/rebar3_hex_utils.erl
```diff
@@ -53,27 +53,54 @@ format_error({not_valid_repo, RepoName}) ->
...
-            Port = open_port({spawn, "tty_sl -e"}, [binary, eof]),
-            port_command(Port, <<?OP_PUTC, Msg/binary>>),
-            receive
-                {Port, {data, PwLine}} ->
-                    [Pw | _] = binary:split(PwLine, <<"\n">>),
-                    port_command(Port, <<?OP_PUTC, $\n>>),
-                    port_close(Port),
-                    error_logger:tty(true),
-                    Pw
+            case io:setopts([binary, {echo, false}]) of
```
tsloughter: Is this a new feature or something? Just wondering why in the case of non-win32 we have something new as well.

src/rebar3_hex_utils.erl
starbelly: Should be yes

src/rebar3_hex_utils.erl
starbelly: Odd... it doesn't show up like this locally, just in the diff... all is pushed too...

```
 77 get_password(Msg) ->
 78     case os:type() of
 79         {win32, nt} ->
 80             get_win32_password(Msg);
 81         _ ->
 82             case io:setopts([binary, {echo, false}]) of
 83                 ok ->
 84                 PwLine = io:get_line(Msg),
 85                 ok = io:setopts([binary, {echo, true}]),
 86                 io:format("\n"),
 87                 [Pw | _] = binary:split(PwLine, <<"\n">>),
 88                 Pw;
 89             _ ->
 90                 error_logger:tty(false),
 91                 Port = open_port({spawn, "tty_sl -e"}, [binary, eof]),
 92                 port_command(Port, <<?OP_PUTC, Msg/binary>>),
 93                 receive
 94                     {Port, {data, PwLine}} ->
 95                         [Pw | _] = binary:split(PwLine, <<"\n">>),
 96                         port_command(Port, <<?OP_PUTC, $\n>>),
 97                         port_close(Port),
 98                         error_logger:tty(true),
 99                         Pw
100                 end
101             end
102     end.
```

tsloughter: Did you use tabs instead of spaces?

starbelly: Nah, but just forced pushed.

starbelly: Ya know maybe, the second case here should be wrapped in a function, maybe getting a little too unwieldy.

PR #88, opened 2019-01-11
Initial work on hex utils suite

starbelly:
- Break tests up into proper suites
  - Test pretty_print_status/1 and repo/2 in rebar3_hex_utils


starbelly:
@tsloughter this one is ready... I'm gonna swing back to do test the functions that interact with io. Want to hit as much as possible before mucking around with that again.

test/rebar3_hex_publish_SUITE.erl
```diff
@@ -0,0 +1,127 @@
...
+    Repo = maps:merge(hex_core:default_config(),
+			#{name => <<"foo">>, repo => <<"foo">>,
+			  api_url => <<"http://127.0.0.1:3000">>,
+			  repo_url => <<"http://127.0.0.1:3000">>,
+			  repo_verify => false}),
+    State = State = rebar_state:new(),
+    Exp = {error,
+            {rebar3_hex_publish,
+             [{invalid_semver,<<"bad_ver">>,"0.a.1b..0.-foo"}]}},
+    Exp = rebar3_hex_publish:publish(App, Repo, State),
```
tsloughter: You can get assert macros from `-include_lib("stdlib/include/assert.hrl").` and use `?assertMatch(A,B)` instead and then rebar3 is able to format any badmatch in a more readable error during test runs.

starbelly: aye, I'll revise... I got in a bad habit of doing this.

starbelly: aight, fixed that up and added a few more test suites (start of them anyway) while we're here.

starbelly: Right now we're pulling in the assertions via eunit.hrl, I did that as that's how you had it in the original suite. Keep like this or change to just pull in the assert.hrl?

tsloughter: Eh, either is fine.

tsloughter (approved):

PR #90, opened 2019-01-18
chomp password in get_win32_password

starbelly:


jkrukoff:
I think this'll fix the problem, though it does look like slightly different semantics than the method the non-windows password entry is using:

```
            [Pw | _] = binary:split(PwLine, <<"\n">>),
```

starbelly:
@jkrukoff that would work too, but easiest to chomp it will it's a char list IMO

PR #94, opened 2019-01-19
Ease test coverage across multiple OTP releases

g-andrade:
* Support Travis CI integration
* Run test cases under all OTP minor releases from OTP 18.0 to OTP 21.2

(This will only have an effect if Travis CI becomes aware of the repo, which I've done in my fork only, as to make changes in #93 easier to guide and prevent future breakage.)


tsloughter:
I'd prefer we kept it using circleci. I'll try to get that done soon and will leave this open until it is.

paulo-ferraz-oliveira:
Should we consider GitHub Actions as a viable alternative? I'm OK to help with this one, but know nothing about CircleCI.

tsloughter:
Definitely should move to Github Actions. Surprised it wasn't already...

starbelly:
Agreed. +1 on github actions.

PR #95, opened 2019-01-21
Build matrix and dialyzer fixes

starbelly:


starbelly:
@tsloughter Could maybe add some more versions to test against, but other than that  this is ready to go.

.circleci/config.yml
```diff
@@ -1,10 +1,125 @@
...
+
+                - store_test_results:
+                    path: ~/project/_build/test/logs/
+                - store_artifacts:
+                    path: ~/project/_build/test/logs
+                    destination: common_test
+
+                - rebar3/cover
+                - run: rebar3 as test covertool generate
+                # - run:
```
tsloughter: Why is this commented out?

starbelly: Optimally it shouldn't be, let's go ahead and uncomment and add some badges while we're here too, yeah?

rebar.config
```diff
@@ -6,11 +6,37 @@
...
-        {deps, [{erlware_commons, "1.2.0"}, {elli, "3.0.0"}, {jsone, "1.4.7"}, {meck, "0.8.12"}]},
+        {overrides, [{override, rebar3,[{deps, [{erlware_commons, "1.2.0"}]}]}]},
+        {deps, [{hex_core, "0.4.0"}, {rebar, {git, "git://github.com/erlang/rebar3.git", {tag, "3.8.0"}}},
+                {erlware_commons, "1.2.0"}, {elli, "3.0.0"},
+                {jsone, "1.4.7"}, {meck, "0.8.12"}]},
         {erl_opts, [nowarn_export_all]}
     ]}
 ]}.
 
+{xref_ignores, [
```
tsloughter: Just set the xref checks to:

```
{xref_checks,[undefined_function_calls,undefined_functions,locals_not_used,
              deprecated_function_calls,deprecated_functions]}.
```

instead of listing all these ignores.

starbelly: Yeah, I was trying to shoot for specificity (i.e., only ignoring what should be ignored), but I suppose that's most of everything.

.circleci/config.yml
tsloughter: That or just remove it from this PR and we can worry about it later.

tsloughter:
I don't think we need to test against so many OTPs. How about just the latest release of each? Which I think the images without a full version number are the latest for those major versions, so we'd just need 21, 20, 19, 18

starbelly:
That works for me.

tsloughter (approved):

PR #99, opened 2019-02-07
fix syntax errors in get_win32_password

starbelly:
- receive patterns require a return
  - functions must end with `end`
  - must use length/1 vs size/1 on strings


starbelly:
Additionally, is it possible with circle-ci to not allow a merge if the build is failing?

tsloughter:
@starbelly that isn't a circleci thing, it is github. It should be working just as it does with travis where it requires override the failure to merge.

PR #101, opened 2019-02-11
Add gitignore pattern to address rouge file regression

starbelly:
Ignore all files in src that do not match src/rebar3_hex.*\.(erl|hrl|.app.src) via gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,8 @@ _plugins
 _tdeps
 logs
 doc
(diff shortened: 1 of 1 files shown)


tsloughter:
hm? I don't think we need this. Just don't add files that shouldn't be there :)

starbelly:
Yeah, that's a pretty good rule 🤔

PR #106, opened 2019-02-27
Honor the -r switch when publishing with other minor fixups/enhancements

starbelly:
- use hexpm as the default vs an org if no repo arg was passed
  - use hexpm as repo in tests
  - add publish org test and publish org error test
  - update reset_password to expect a 204 vs a 201 and binarify username
  - ensure hex_core 0.5.0 is used in tests
  - add route for org publishing
  - remove repo option from rebar3_hex_user
  - search should return state on no results
  - use starbelly/rebar3.git until #2020 is merged
  - tableize search results
  - added print_table/1 to rebar3_hex_utils
  - added verl to deps for sussing out stable releases
  - Added rebar3_hex_utils:str_split/2 for compat string splitting for  OTP versions prior to 20


starbelly:
This should have been included in the PR from yesterday.

starbelly:
This must remain a WIP until https://github.com/erlang/rebar3/pull/2020 is merged, if we don't want my repo in the rebar.config for the test profile that is.

PR #107, opened 2019-03-22
Fix win32 password entry.

jkrukoff:
As `Prompt` is a binary, this patch prevents a run time error on
password blanking. This appears to be the only change needed to get the plugin working on windows again, after the original botched merge and subsequent emergency fix.


starbelly:
@jkrukoff yeah, i changed it to size because for some reason I got back strings and not binary, but it's been a minute, so I may not be remembering something correctly.

PR #112, opened 2019-04-15
Use first repository organization over default

starbelly:
In the case where the user has a repo in their config: 

```
{erl_opts, [debug_info]}.
{deps, [rebar3_hex]}.
{hex, [{repos, [#{name => <<"hexpm:jellybeam">>}]}]}.
```

And the user doesn't supply the repo the want to publish to on the command line, accidents can happen (i.e., you publish your private package to the public hexpm repo). In order to mitigate this we should follow suite with mix/hex and default to using the repo defined in config vs the default repo... if no repo is found in config, then prefer the default. 

 Now, iirc the intention was to allow a project to have multiple repositories defined in config. That part I still don't grok as you can not publish a private package to an org with dependencies from other orgs per the hex docs. If we still want that flexibility, then we'll want to provide an additional configuration option such as `organization`. I still don't grok the multi-repo bit, but yeah. It may be that I am not recalling correctly and that we only put repos in a list for the sake of merging with the default repo. 

 Regardless, if the user wants to be able to avoid accidents, they should be able to do so. The `-r` switch would still be honored of course and take precedence over what's in the config file.

### DIFF
--- a/src/rebar3_hex_utils.erl
+++ b/src/rebar3_hex_utils.erl
@@ -42,18 +42,28 @@ pretty_print_errors(Errors) ->
 repo_opt() ->
     {repo, $r, "repo", string, "Repository to use for this command."}.
 
+get_first_repo_org_or_default(Repos) ->
+   case get_first_repo_org(Repos) of
+       [Repo] ->
+           {ok, Repo};
+       [] ->
+           {ok, rebar_utils:to_binary(?DEFAULT_HEX_REPO)}
+   end.
+
+get_first_repo_org(Repos) ->
+   lists:filter(fun(X) ->
+                         Name = maps:get(name, X, undefined),
+                         Name  =/= <<"hexpm">> andalso Name  =/= undefined
+                 end, Repos).
+
 repo(State) ->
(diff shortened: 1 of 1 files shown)


src/rebar3_hex_publish.erl
```diff
@@ -48,6 +48,7 @@ init(State) ->
 do(State) ->
     Apps = rebar3_hex_utils:select_apps(rebar_state:project_apps(State)),
     Repo = rebar3_hex_utils:repo(State),
+    erlang:display(Repo),
```
tsloughter:
```suggestion
```

tsloughter:
Maybe we just shouldn't allow publishing without passing an explicit repo if there are multiple repos defined.

starbelly:
@tsloughter Either way is fine. I think we currently do what we do do be in line with hex UX, but I am ok with either. Whatever is going to provide more "safety".

tsloughter:
Yea, I think safest would be to require `-r` unless there are no custom repos defined.

starbelly:
K, I'll open a new PR to do that sometime tomorrow.

PR #114, opened 2019-05-26
Make r switch mandatory when non-default repos are detected for state changing ops

starbelly:
This PR aims to provide a consistent API when deciding what commands require a repo argument and more specifically when.

Things in this PR not directly related to the above:

    - move search result functions back into rebar3_hex_search, they never belonged in the utils module.
    - added maybe_throws/2 helper on utils module for handling function that maybe throw (specifically for breaking out of recursive functions). 

To recap on the API part here are the rules for when a repo argument is required going forward:

# Cut
 - **cut** requires a repo arg if non-defaults exist

# Docs
 - **docs** requires a repo arg if non-defaults exist

# Key
 - **list** does not require a repo arg, optionally can take one, otherwise aggregates results
 - **remove** requires a repo arg if non-defaults exist

# Owner
 - **add** requires a repo arg if non-defaults exist
 - **remove** requires a repo arg if non-defaults exist
 - **list** does not require a repo arg, can optionally take one, otherwise aggregates results

# Publish
 - **publish** requires a repo a arg if non-defaults exist

# Repo 
 - **generate** always requires a repo arg
 - **auth** - always requires a repo arg

# Retire
 - **retire** requires a repo arg if non-defaults exist

# Revert
 - **revert** requires a repo arg if non-defaults exist

# User
 - **register** requires a repo arg if non-defaults exist
 - **auth** requires a repo arg if non-defaults exist
 -  **deauth** requires a repo arg if non-defaults exist
 - **reset_password** requires a repo arg if non-defaults exist
 - **whoami** does not require a repo arg, can optionally take one, otherwise aggregates results

# Search
 - **search** does not require a repo arg, can optionally take one, otherwise aggregates results

### DIFF
--- a/src/rebar3_hex_cut.erl
+++ b/src/rebar3_hex_cut.erl
@@ -35,7 +35,14 @@ init(State) ->
 
 -spec do(rebar_state:t()) -> {ok, rebar_state:t()} | {error, string()}.
 do(State) ->
-    Repo = rebar3_hex_utils:repo(State),
+    case rebar3_hex_utils:repo(State) of
+        {ok, Repo} ->
+            handle_command(State, Repo);
+        {error, Reason} ->
+            ?PRV_ERROR(Reason)
+    end.
+
+handle_command(State, Repo) ->
     case maps:get(write_key, Repo, undefined) of
         undefined ->
             ?PRV_ERROR({no_write_key, maps:get(name, Repo)});
--- a/src/rebar3_hex_docs.erl
+++ b/src/rebar3_hex_docs.erl
@@ -26,7 +26,6 @@ init(State) ->
                                 {example, "rebar3 hex docs"},
                                 {short_desc, "Publish documentation for the current project and version"},
                                 {desc, ""},
-
                                 {opts, [{revert, undefined, "revert", string, "Revert given version."},
                                         rebar3_hex_utils:repo_opt()]},
                                 {profiles, [docs]}]),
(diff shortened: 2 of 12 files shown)


tsloughter:
Yes, it should be required for `cut` as well.

starbelly:
Ok, got some feedback from Fred... so, what we can do is require a repo name for all hex ops that require a repo when non-defaults exist. So, in the case of not having any non-default configs, we just publish (i.e., no repo name is required). If non-default configs are found and no name was provided, bail. 

Sound good?

tsloughter:
yup

starbelly:
@tsloughter hmm, look at the latest changes. We will require supplying the repo for searching... maybe it shouldn't be this way for that... though, a valid read key is indeed required for searching 🤔 

 We could add an optional param for rebar3_hex_utils:repo/1... `use_first` where the first repo in config would be used. This creates a problem though if the intention is to allow searching repositories other than hexpm.

starbelly:
To recap, here are the rules for when a repo argument is required going forward:

# Cut
 - **cut** requires a repo arg if non-defaults exist

# Docs
 - **docs** requires a repo arg if non-defaults exist

# Key
 - **list** does not require a repo arg, optionally can take one, otherwise aggregates results
 - **remove** requires a repo arg if non-defaults exist

# Owner
 - **add** requires a repo arg if non-defaults exist
 - **remove** requires a repo arg if non-defaults exist
 - **list** does not require a repo arg, can optionally take one, otherwise aggregates results

# Publish
 - **publish** requires a repo a arg if non-defaults exist

# Repo 
 - **generate** always requires a repo arg
 - **auth** - always requires a repo arg

# Retire
 - **retire** requires a repo arg if non-defaults exist

# Revert
 - **revert** requires a repo arg if non-defaults exist

# User
 - **register** requires a repo arg if non-defaults exist
 - **auth** requires a repo arg if non-defaults exist
 -  **deauth** requires a repo arg if non-defaults exist
 - **reset_password** requires a repo arg if non-defaults exist
 - **whoami** does not require a repo arg, can optionally take one, otherwise aggregates results

# Search
 - **search** does not require a repo arg, can optionally take one, otherwise aggregates results

PR #128, opened 2019-07-08
Always store an .app.src file in the package

dumbbell:
In pull request #116, the behavior changed a bit: the generated `.app.src` file was included in the package (the tar archive) only if one was already listed in the included files. Before this changed, the generated `.app.src` file was already prepended to that list.

Therefore, if the initial `.app.src` was missing from that list, no `.app.src` was added to the package.

This commit restores the previous behavior: the generated `.app.src` replaces the initial file or it is added (appended) to the list.

This solves an issue where Erlang.mk-based packages (which may not have an `.app.src` file on disk), published using this plugin to Hex.pm, were published successfully but were considered invalid when fetched as a dependency later:

```
===> Dependency failure: source for (...) does not contain a recognizable project and can not be built
```

### DIFF
--- a/src/rebar3_hex_publish.erl
+++ b/src/rebar3_hex_publish.erl
@@ -272,7 +272,7 @@ include_files(Name, AppDir, AppDetails) ->
 
     AppFileSrc = filename:join("src", ec_cnv:to_list(Name)++".app.src"),
     AppSrcBinary = ec_cnv:to_binary(lists:flatten(io_lib:format("~tp.\n", [AppSrc]))),
-    lists:keyreplace(AppFileSrc, 1, WithIncludes, {AppFileSrc, AppSrcBinary}).
+    lists:keystore(AppFileSrc, 1, WithIncludes, {AppFileSrc, AppSrcBinary}).
 
 
 is_valid_app({_App, _Name, _Version, _AppDetails} = A) ->


tsloughter:
@dumbbell oh, does rebar3_hex generate a .app.src file for erlang.mk files from the .app file? I guess this is fine, I'm just surprised it works :)

dumbbell:
Yes. In fact, Erlang.mk does not provide a way to publish to Hex.pm out-of-the-box. So we developped a small tool (https://github.com/rabbitmq/hexpm-cli) to be able to use the `rebar3_hex` plugin outside of a full-blown Rebar 3 project. That escript embeds Rebar with a couple functions patched. This includes the path to the `.app.src` file which points to the generated `.app` file.

dumbbell:
Thank you for merging this patch!

Could you please publish a new release of the plugin?

tsloughter:
Can't you just use erlang.mk's support for generating a .app file in ebin?

I know it is more work than rabbit wants to take on when they spent the time moving to erlang.mk, and I'm sure you already thought of this and decided it was less work than supporting rebar3, but have to mention anyway that this will likely break at times. It looks small so probably less effort to keep it working than to support rebar3, just felt I should mention it :)

dumbbell:
We use Erlang.mk's ability to generate the `.app` file. Btu when I tested, it was not enough to use Rebar out-of-the-box. Unfortunately, I don't remember what was missing or not working, leading to that tool embedding Rebar.

dumbbell:
@tsloughter: Would it be possible to release a new version of the plugin with this fix? Is there anything I can do to help with this?

I'm asking because we can't publish new versions of the RabbitMQ Erlang client to Hex.pm currently.

Thank you very much!

tsloughter:
@dumbbell yes, I want to test and merge #129 first. I will do that this morning and then cut a release.

dumbbell:
Awesome! Thank you very much :-)

tsloughter:
Published 6.8.0.

PR #129, opened 2019-07-10
Avoid failing when the .app.src contains git or semver as version

jfacorro:


tsloughter:
I thought this would already work, though I must confess I almost always use `rebar3 hex cut`. Is there a failure you hit you can describe?

jfacorro:
I've experienced running `rebar3 hex publish` fails when using it in a project where the `.app.src` contains the atom `git` as the value for the `version` entry, the command fails with:
```
===> Using deprecated rebar_utils:vcs_vsn/3. Please upgrade your plugins.
===> Getting log of git repo failed in /Users/juan.facorro/foo. Falling back to version 0.0.0
```
Steps to reproduce it:
```
$ rebar3 new app foo
$ cd foo
$ vim src/foo.app.src
# Change the value of 'version' to the atom 'git'
$ rebar3 hex publish
```
The following is the result of running in `DEBUG`:
```
$ DEBUG=1 rebar3 hex publish
===> Load global config file /Users/juan.facorro/.config/rebar3/rebar.config
===> 20.3.8.21 satisfies the requirement for minimum OTP version 18
===> Expanded command sequence to be run: []
===> Provider: {default,do}
===> Expanded command sequence to be run: [{default,app_discovery},
                                                  {default,install_deps},
                                                  {default,lock},
                                                  {hex,publish}]
===> Provider: {default,app_discovery}
===> Provider: {default,install_deps}
===> Provider: {default,lock}
===> Provider: {hex,publish}
===> Using deprecated rebar_utils:vcs_vsn/3. Please upgrade your plugins.
===> sh info:
        cwd: "/Users/juan.facorro/foo"
        cmd: git log -n 1 --pretty=format:"%h
"

===>    opts: [{use_stdout,false},
                       return_on_error,
                       {cd,"/Users/juan.facorro/foo"}]

===> Port Cmd: git log -n 1 --pretty=format:"%h
"
Port Opts: [{cd,"/Users/juan.facorro/foo"},
            exit_status,
            {line,16384},
            use_stdio,stderr_to_stdout,hide,eof]

===> Getting log of git repo failed in /Users/juan.facorro/foo. Falling back to version 0.0.0
===> Uncaught error: function_clause
===> Stack trace to the error location:
[{ec_semver_parser,parse,
                   [git],
                   [{file,"/Users/juan.facorro/dev/jfacorro/rebar3/_build/default/lib/erlware_commons/src/ec_semver_parser.erl"},
                    {line,20}]},
 {rebar3_hex_publish,validate_app,2,
                     [{file,"/Users/juan.facorro/.cache/rebar3/plugins/rebar3_hex/src/rebar3_hex_publish.erl"},
                      {line,295}]},
 {rebar3_hex_publish,'-is_valid_app/1-fun-0-',3,
                     [{file,"/Users/juan.facorro/.cache/rebar3/plugins/rebar3_hex/src/rebar3_hex_publish.erl"},
                      {line,280}]},
 {lists,foldl,3,[{file,"lists.erl"},{line,1263}]},
 {rebar3_hex_publish,is_valid_app,1,
                     [{file,"/Users/juan.facorro/.cache/rebar3/plugins/rebar3_hex/src/rebar3_hex_publish.erl"},
                      {line,287}]},
 {rebar3_hex_publish,publish,3,
                     [{file,"/Users/juan.facorro/.cache/rebar3/plugins/rebar3_hex/src/rebar3_hex_publish.erl"},
                      {line,141}]},
 {lists,foldl,3,[{file,"lists.erl"},{line,1263}]},
 {rebar_core,do,2,
             [{file,"/Users/juan.facorro/dev/jfacorro/rebar3/src/rebar_core.erl"},
              {line,154}]}]
```

tsloughter:
In the example you give there isn't a git repo. Did you just miss a step in the reproduction or is that related?

jfacorro:
I didn't include that part (i.e. doing `git init`, `git add .`, etc.) because the result is the same if the project is in a git repo. 😃

The error is cause by the fact that the value provided to `is_valid_app/1` is the one in the `.app.src` and not the one resolved by `rebar3`.

starbelly:
@jfacorro Can you try the latest version of rebar3_hex? The semver function from erlware_commons was replaced by verl in later versions. Note that this dictates your version be a valid semver.

starbelly:
Ahh my bad, I'm wrong. We're only using verl in search right now. @tsloughter Perhaps we should switch to verl for the validation at this point though.

tsloughter:
I want to try this before merging, will probably get around to it today.

tsloughter:
Hm, so I discovered rebar3_hex already works but only if you use `"git"` and not `git` atom.. So there may be a better fix for this.

tsloughter:
Oooooh, I see what is going on.

I just realized the patch fixes what gets sent to @starbelly 's new `has_semver` check :). So the version is being set correctly, just not when that is called.

tsloughter:
What got me so confused is the changes you did to `rebar_utils:vcs_vsn`, I don't think those are needed.

I'll comment with a suggested patch.

src/rebar3_hex_publish.erl
```diff
@@ -126,9 +126,7 @@ publish(App, HexConfig, State) ->
     Name = rebar_app_info:name(App),
 
     Version = rebar_app_info:original_vsn(App),
-    ResolvedVersion = rebar_utils:vcs_vsn(Version,
-                                          rebar_app_info:dir(App),
-                                          rebar_state:resources(State)),
+    ResolvedVersion = rebar_utils:vcs_vsn(App, Version, State),
```
tsloughter:
```suggestion
    ResolvedVersion = rebar_utils:vcs_vsn(Version,
                                          rebar_app_info:dir(App),
                                          rebar_state:resources(State)),    
```

PR #130, opened 2019-07-14
Use the win32 method for getting passwords on all platforms

starbelly:
In light of `io` problems that should be dealt with in Erlang/OTP in regards to echoless prompts we should switch to the same method of prompting for passwords that we use on win32 ala mix/hex.

Specifically, we are seeing issues where by `io:setops/1` now consistently returns `{error, notsup}` where it should return `ok` per the platform (e.g., mac os, linux, *bsd, etc.). Until we can safely depend on `io:getline/1` and/or `io:get_password/1` we should resort to what appears to be the only reliable method for getting a password. That is reading and clearing the input at a prompt every 1ms.

### DIFF
--- a/src/rebar3_hex_utils.erl
+++ b/src/rebar3_hex_utils.erl
@@ -136,40 +136,10 @@ format_error({not_valid_repo, RepoName}) ->
 
 -define(OP_PUTC, 0).
 
-get_password(Msg) ->
-    case os:type() of
-        {win32, nt} ->
-            get_win32_password(Msg);
-        _ ->
-            get_tty_password(Msg)
-    end.
-
-get_tty_password(Msg) ->
-    case io:setopts([binary, {echo, false}]) of
-        ok ->
-            PwLine = io:get_line(Msg),
-            ok = io:setopts([binary, {echo, true}]),
-            io:format("\n"),
-            [Pw | _] = binary:split(PwLine, <<"\n">>),
-            Pw;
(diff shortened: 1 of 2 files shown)


tsloughter:
I'll just note that we also saw issues with the `port_command` version of the password prompt that it uses when `io:setopts/1` returns an error. This would successfully get the password with a blank prompt but it would cause any prompt in the program after that to not accept input.

So in the case of the `cut` command after getting the local password to decode the write key it asks if you want to push the new tag to the remote git repo and would appear to hang when you try to input anything.

ferd:
is the `enotsup` error returned based on whether the old-style `user` process or the new one? Or it happens regardless?

tsloughter:
old-style? Like the old shell? It does for either new or old shell.

ferd:
yeah that was the question. If it works for neither then it's not good.

tsloughter:
Yea, neither.

starbelly:
Realized what changed. This bug had laid dormant. Prior to 7f72a34e4959dfab3cb299910bd06c266203c200 the order of prompts in rebar3_hex_cut didn't allow for this bug to be experienced. Specifically, asking for the password was the final prompt. 

 With that in mind, do we want to restore the old order of prompts for this command and revert this patch? Might be counter-productive for moving forward.

tsloughter:
No, we should leave this as it is the only one that works fully.

PR #133, opened 2019-08-04
Remove rebar3_hex_utils and reorganize per removal

starbelly:
This PR is part of an "epic" to reorganize, cleanup, improve tests, add docs and support organization transfers. There are at least 4 PRs to follow. As such, a release should not be made until the "epic" is complete.

 This PR : 
 - Removes rebar3_hex_utils and re-organizes the functions that lived in that module. 
 - Provides a few minor fixups and/or enhancements, namely in the form of rebar3_hex_errors as a "catch all" that command modules can call out to in the case of an error that is unknown. 
 - Removes erlware_commons dependency by introducing rebar3_hex_io module


src/rebar3_hex_file.erl
```diff
@@ -0,0 +1,35 @@
...
+
+-export([
+         expand_paths/2,
+         update_app_src/2
+        ]).
+
+expand_paths(Paths, Dir) ->
+    AbsDir = filename:absname(Dir),
+    Files = lists:flatmap(fun dir_files1/1, [filename:join(Dir, P) || P <- Paths]),
+    [{relative_path(F1, AbsDir), F1} || F1 <- filter_regular(Files)].
```
starbelly, line 11: @ferd Pointing you to this since the PR you commented on is actually the bulk of this one until it gets merged. It appears rebar_dir already provides us with a tried and true approach for making a relative path.

starbelly:
Closing to combine this with #134

PR #134, opened 2019-08-17
Publishing enhancements and refactors

starbelly:
- Moved get_required/2, task_args/1, and repo_opt/0 from
       rebar3_hex_utils into rebar3_hex
     - Added rebar3_hex_io
     - Moved io related functions from rebar3_hex_utils into rebar3_hex_io
     - Removed erlware_commons depedency per the new rebar3_hex_io module,
       verl and rebar_utils
     - Added rebar3_hex_config
     - Moved config related functions from rebar3_hex_utils into rebar3_hex_config
     - Added rebar3_hex_file
     - Moved file and path related functions into rebar3_hex_file
     - Removed rebar3_hex_utils and test suite data
     - Added rebar3_hex_error for common error formatting
     - Updated tests where required
     - default to publishing docs when publishing a package
     - improve error flow
     - add tests


src/rebar3_hex_file.erl
```diff
@@ -0,0 +1,32 @@
...
+
+-export([
+         expand_paths/2,
+         update_app_src/2
+        ]).
+
+expand_paths(Paths, Dir) ->
+    AbsDir = filename:absname(Dir),
+    Files = lists:flatmap(fun dir_files1/1, [filename:join(Dir, P) || P <- Paths]),
+    [{F1 -- (AbsDir++"/"), F1} || F1 <- filter_regular(Files)].
```
ferd: That `F1 -- (AbsDir++"/")` seems mighty unsafe. See for example:

```
1> "my-baby-bacon/file" -- "/a/b/c/file".
"my-by-baon"
2> "/a/b/c/d" -- "C:\\\\a\\b\\c\\d".      
"////"
```
You have to be damn sure every single one of the file paths are actually the same. It is much much safer to use `filename:split/1` and iterate over all fragments. See the following functions from rebar_utils as an example of how to do it: https://github.com/erlang/rebar3/blob/5bc4581d6c7336790ddb51e4dce63bc47f8d2754/src/rebar_file_utils.erl#L430-L455

starbelly: Yes, I did not write this function, I just carried it over. Looks like rebar_utils provides us with what we need and if so, we should just use that.

starbelly: Resolving this for https://github.com/tsloughter/rebar3_hex/pull/133/commits/f27ec8f92b16ba80aef3f28aed2e3c0102996b79#r315303574

PR #137, opened 2019-12-25
bump hex_core to 0.6.4

starbelly:
- Bump hex_core to 0.6.4
  - Bump verl to 1.0.2
  - Updated rebar.config and rebar.lock
  - Updated source and tests where applicable per non-backwards compat changes in
    hex_core
  - expose ownership level and transfer options on add for rebar3_hex_owner
  - handle 500s on reset password for rebar3_hex_user


starbelly:
@tsloughter This should be good to merge now.

PR #139, opened 2019-12-29
add support for CI friendly publishing

starbelly:
This allows a user to circumvent traditional config files and prompts suitable for automating CI workflows. 

Allows for the following: 

```
HEX_API_KEY=some_write_key rebar3 hex publish --yes
```

Look ma, no prompts!

### DIFF
--- a/src/rebar3_hex_config.erl
+++ b/src/rebar3_hex_config.erl
@@ -45,12 +45,52 @@ repo(State, RepoName) ->
     MaybeFound2 =  get_repo(<<MaybeParentRepo/binary, BinName/binary>>, Repos),
     case {MaybeFound1, MaybeFound2} of
         {{ok, Repo1}, undefined} ->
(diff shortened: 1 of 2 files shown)


starbelly:
One last thing to add a `--yes` option.

starbelly:
This one is good to go 🎅

tsloughter:
I thought mix/hex were explicitly against a feature like this. But mix actually supports it?

starbelly:
@tsloughter mix supports it https://hexdocs.pm/hex/Mix.Tasks.Hex.Config.html#content . See config overrides. 

  I thought about this myself a bit and came to the conclusion that people will inevitability automate this with or without this support. So, the question then becomes whether to discourage it in documentation or not, but that's another story.

starbelly:
It should be noted mix also provides a `--yes` option on the publish task.

PR #140, opened 2020-01-16
publish package response doesn't have content-type response header

f3c0:
after runing  `$ rebar3 hex docs` I got the `===> {publish,{created,nil}}` result
the request was success, but was not `{created, <<>>}` the result

### DIFF
--- a/src/rebar3_hex_docs.erl
+++ b/src/rebar3_hex_docs.erl
@@ -91,7 +91,7 @@ do_publish(App, State, Repo) ->
     {ok, Config} = rebar3_hex_config:hex_config_write(Repo),
 
     case rebar3_hex_client:publish_docs(Config, rebar_utils:to_binary(PkgName), rebar_utils:to_binary(Vsn), Tar) of
-        {created, <<>>} ->
+        {created, _} ->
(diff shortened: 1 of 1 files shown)


starbelly:
Hi @f3c0! Great catch! 

 Right now I'm trying to suss out whether this should be `nil` or `undefined` in hex_core via @ericmj. The change happened in https://github.com/hexpm/hex_core/commit/1155a06b49289bbc09ac04663d38f003e49258e2 but clearly their wasn't enough integration testing done on our end which would have caught this prior to the last release of rebar3_hex.  
  
 That said I am happy to merge this and do a patch release. 

@tsloughter thoughts?

f3c0:
I'm not very familiar with the hex api, but it looks it returns nil always if the header content type is not the expected erlang type even if it has body or not.
For a general api client I would not hide the body from the user. And in the plug-in I would match to {created, _} instead of {created, nil}

starbelly:
@f3c0 Yes, it used to return `<<>>`, but fair enough, can you modify your PR to match on `{created, _}` ?

f3c0:
done

starbelly:
Thank you! ❤️

PR #151, opened 2020-03-18
Handle key list with repo argument

starbelly:
Due to how we were matching on command arguments a key list command with a repo argument would result a in crash.

 - add rebar3_hex:sub_command/1
 - refactor handle_command/2 to handle_command/3 in rebar3_hex_key
 - update test deps in rebar.config

### DIFF
--- a/src/rebar3_hex.erl
+++ b/src/rebar3_hex.erl
@@ -1,6 +1,6 @@
 -module(rebar3_hex).
 
--export([init/1, get_required/2, task_args/1, repo_opt/0, help_opt/0, sub_command/1]).
+-export([init/1, get_required/2, task_args/1, repo_opt/0, help_opt/0]).
 
 init(State) ->
     lists:foldl(fun provider_init/2, {ok, State}, [rebar3_hex_user,
--- a/src/rebar3_hex.erl
+++ b/src/rebar3_hex.erl
@@ -1,6 +1,6 @@
 -module(rebar3_hex).
 
--export([init/1, get_required/2, task_args/1, repo_opt/0, help_opt/0]).
+-export([init/1, get_required/2, task_args/1, repo_opt/0, help_opt/0, sub_command/1]).
 
 init(State) ->
     lists:foldl(fun provider_init/2, {ok, State}, [rebar3_hex_user,
@@ -28,6 +28,12 @@ get_required(Key, Args) ->
 task_args(State) ->
     {[{task, Task} | Args], _} = rebar_state:command_parsed_args(State),
     {Task, Args}.
+
+sub_command(State) ->
+  case rebar_state:command_args(State) of
+      [] -> undefined;
+      [Cmd|_] -> Cmd
+  end.
 
 repo_opt() ->
   {repo, $r, "repo", string, "Repository to use for this command."}.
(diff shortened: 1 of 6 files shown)


src/rebar3_hex.erl
```diff
@@ -29,6 +29,12 @@ task_args(State) ->
     {[{task, Task} | Args], _} = rebar_state:command_parsed_args(State),
     {Task, Args}.
 
+sub_command(State) ->
```
starbelly: I'm not terribly sure this is the best way to skin this cat, I don't see harm in it either. As a counter example, in rebar3_hex_repo we make use of `{Args, _} = rebar_state:command_parsed_args(State)` and `proplists:get_value(subcmd, Args, undefined)` but pretty sure they both do the same thing, maybe there's added checks and balances from rebar3 if we get the subcmd from the `subcmd` tuple though.

starbelly: Ignore the above, `subcmd` is not reliable. However, I'm thinking about doing away with "lose arguments". When we use switches we can get all the arguments in a normalized form, lose arguments (such as the key name argument to the revoke sub command) are well, lose. We end up having to have special cases for these and we rely that they are specified in exactly the right order. 

So, instead of `rebar3 hex key revoke bar` we would just switch to `rebar3 hex key revoke --key-name bar` or `rebar3 hex key revoke -k bar`. The latter is less ergonomic, but I think the trade off is worth it.

starbelly: I'm going to push this change up and we can decide to keep it or not.

src/rebar3_hex_error.erl
```diff
@@ -4,6 +4,22 @@
...
 format_error({required, repo}) ->
     "A repository argument is required for this command.";
+
+format_error({error, no_read_key}) ->
+    "No read key found for user. Be sure to authenticate first with:"
+    ++ " rebar3 hex user auth";
+
+format_error({error, no_write_key}) ->
+    "No write key found for user. Be sure to authenticate first with:"
+    ++ " rebar3 hex user auth";
```
ferd: You don't need to use `++` between two literal strings, they get auto-merged

src/rebar3_hex_key.erl
```diff
@@ -1,141 +1,154 @@
...
 init(State) ->
     Provider = providers:create([{name, ?PROVIDER},
                                  {module, ?MODULE},
                                  {namespace, hex},
                                  {bare, true},
                                  {deps, ?DEPS},
-                                 {example, "rebar3 hex key [generate <key> | list | revoke <key> | revoke --all]"},
-                                 {short_desc, "Remove or list API keys associated with your account"},
+                                 {example,
+                                  "rebar3 hex key [generate -k <key> | list | revoke -k <key> "
```
ferd, line 18: this should probably come with a major version bump? Or is this considered a bug fix?

ferd (approved):
Approved but check the question about version bumps

src/rebar3_hex_key.erl
starbelly: Two bug fixes... one in the handling of the args and command, second in the documentation.

PR #153, opened 2020-04-05
Add catch all for publishing when a message is available

### LINKED ISSUE #150: Publishing using an unauthorized user crashes
Publishing using an unauthorized user crashes:

```
...
Proceed? ("Y")> y
Local Password:
===> Uncaught error: {case_clause,
                                {ok,{403, ...
```

* `rebar3` v3.13.1 on Erlang/OTP 22 Erts 10.6.4
* `rebar3_hex` v6.93
* `hex_core` v0.6.8
* `verl` v1.0.2

starbelly:
- Instead of matching on every possible http status code,
  have a catch all for when there it is an unexpected status but has a
  message from the server.
- closes #150


starbelly:
We should be moving towards using rebar3_hex_client, yet I'd rather refactor for that in another PR, so here is a small patch that deals directly with the issue at hand. 

Manually tested against a hex.pm instance, but need to add a "integration" test to go with this.

starbelly:
Added test, this one is good got review and merge.

ferd (approved):

PR #156, opened 2020-04-13
Fix git docs publish

egobrain:
Fix docs publish in case app.src vsn is set to `git`


starbelly (approved):

starbelly:
This makes sense to me 👍 , would ask @tsloughter to take a look at it too though.

starbelly:
I QAed and found the bug this fixes 👍

PR #167, opened 2020-05-24
Add support for replace package option

starbelly:
hex_core and hex.pm have introduced an option to replace an existing
package, this commit surfaces

- upgrade to hex 0.6.9

- add --replace switch on publish command

- Fix up publish options, previously there were strings where atoms were
  expected (element 0 of the option tuple)

- make --revert fully supported and useable

- add --package to support the --revert switch

- Start follow the command pattern laid out by the key command

- Improve description with supportive documentation

- move call to hex_api_release into rebar3_hex_client as publish/2

- update test suites were required

# Output of rebar3 help for publish command

```
Publishes a new version of a package with options to revert and replace existing packages

Supported commmand combinations:

  rebar3 hex publish

  rebar3 hex publish --yes

  rebar3 hex publish --repo <repo>

  rebar3 hex publish --repo <repo> --yes

  rebar3 hex publish --revert <version> --package <package>

  rebar3 hex publish --revert <version> --package <package> --yes

  rebar3 hex publish --replace

  rebar3 hex publish --replace --yes

Argument descriptions:

  <repo>    - a valid repository, only required when multiple repositories are configured

  <version> - a valid version string, currently only utilized with --revert switch

  <package> - a valid package name, currently only utilized with --revert switch


Usage: rebar3 hex publish [-r <repo>] [-y [<yes>]] [--replace [<replace>]]
                          [-p <package>] [--revert <revert>]

  -r, --repo     Repository to use for this command.
  -y, --yes      Publishes the package without any confirmation prompts
                 [default: false]
  --replace      Allows overwriting an existing package version if it
                 exists. Private packages can always be overwritten,
                 publicpackages can only be overwritten within one hour
                 after they were initially published. [default: false]
  -p, --package  Specifies the package to use with the publish command,
                 currently only utilized in a revert operation
  --revert       Revert given version, if the last version is reverted the
                 package is removed
```

# TODO: 

- [x] Add tests for help on commands to ensure we do not crash
- [x] Add tests for --replace
- [x] Add tests for --revert

### DIFF
--- a/src/rebar3_hex.erl
+++ b/src/rebar3_hex.erl
@@ -1,6 +1,6 @@
 -module(rebar3_hex).
 
--export([init/1, get_required/2, task_args/1, repo_opt/0, help_opt/0]).
+-export([init/1, gather_opts/2, get_required/2, task_args/1, repo_opt/0, help_opt/0]).
 
 init(State) ->
     lists:foldl(fun provider_init/2, {ok, State}, [rebar3_hex_user,
@@ -16,6 +16,17 @@ init(State) ->
 
 provider_init(Module, {ok, State}) ->
     Module:init(State).
+
+gather_opts(Targets, State) -> 
+    {Args, _} = rebar_state:command_parsed_args(State),
+    lists:foldl(fun(T, Acc) -> 
+                        case proplists:get_value(T, Args, undefined) of
+                            undefined -> 
+                                Acc;
+                            V -> 
+                                maps:put(T, V, Acc)
+                        end
+                end, #{}, Targets).
 
 get_required(Key, Args) ->
     case proplists:get_value(Key, Args) of
--- a/test/rebar3_hex_SUITE.erl
+++ b/test/rebar3_hex_SUITE.erl
@@ -5,7 +5,20 @@
 -include_lib("eunit/include/eunit.hrl").
 
 all() ->
-    [init_test, repo_opt].
+    [task_args_test, gather_opts_test, init_test, help_test, repo_opt].
+
+
+gather_opts_test(_Config) -> 
+    State = rebar_state:new(),
+    CmdArgs = {[{foo,"bar"}, {count, 42}, {other, eh}], []},
+    State1 = rebar_state:command_parsed_args(State, CmdArgs),
+    ?assertMatch(#{count := 42, foo := "bar"}, rebar3_hex:gather_opts([count, foo], State1)).
+
+task_args_test(_Config) -> 
+    State = rebar_state:new(),
+    CmdArgs = {[{task, thing} | {[{foo,"bar"}, {count, 42}], []}], []},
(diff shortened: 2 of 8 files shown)


starbelly:
One thing we could do here is support publishing a specific package since we've added a `--package` switch to support `--revert` as a switch. Thoughts?

src/rebar3_hex.erl
```diff
@@ -29,6 +29,18 @@ task_args(State) ->
...
+    {Args, _} = rebar_state:command_parsed_args(State),
+    lists:foldl(fun fold_arg/2, #{}, Args).
+
+fold_arg({K, V}, Acc) ->
+    case maps:get(K, Acc, undefined) of
+        undefined ->
+            maps:put(K, V, Acc);
+        [_|_] = Vals ->
+            maps:put(K, [V | Vals], Acc)
+    end.
```
ferd: this fails if you get duplicate keys and the first key inserted wasn't a list since it won't match either clause. I'm unsure if the values are always meant to be list, crash as implemented here, or meant to be turned to lists when there are more than one

equivalent to: `maps:update_with(K, fun(Vals = [_|_]) -> [V|Vals] end, V, Acc)` right now; `maps:update_with(K, fun(Vals) -> [V|Vals] end, [V], Acc)` for a format that's always lists. For the switch to turning values to lists, it's fundamentally a bit risky since it's not obvious what to do for strings.

starbelly: Eh yeah, good point.

starbelly: Refactored args/1 to gather_opts/2 to be on the safe side of the tracks.

test/rebar3_hex_SUITE.erl
```diff
@@ -14,3 +27,31 @@ repo_opt(_Config) ->
 init_test(_Config) ->
     {ok, State} = rebar3_hex:init(rebar_state:new()),
     ?assertEqual(state_t, element(1, State)).
+
+%% Smoke test to ensure we don't crash
+help_test(_Config) -> 
+    %% Silent output during our tests
```
starbelly, line 33: This is really a brute force method for silencing output... I tried setting the group leader to a dummy proc but had issues... maybe there's a good pattern for this that I'm not aware of?

starbelly:
> One thing we could do here is support publishing a specific package since we've added a `--package` switch to support `--revert` as a switch. Thoughts? 


☝️ ignore this please. Will open another issue or set an agenda item for building & packaging to discuss this further.

tsloughter:
@ferd you reviewed this?

@starbelly is it ready for merge?

I want to publish a release with my `cut` fix so figured it best to get all the PRs that are ready in first.

starbelly:
@tsloughter Yeah, the only thing in question was the hacky way I'm silencing output in a test, but it's no solid reason to hold this up. Let me fix the conflict and merge.

tsloughter (approved):

PR #168, opened 2020-05-25
Fix handling of binary version specs

filmor:


src/rebar3_hex_publish.erl
```diff
@@ -387,6 +387,8 @@ format_build_tools(BuildTools) ->
 update_versions(ConfigDeps, Deps) ->
     [begin
          case lists:keyfind(binary_to_atom(N, utf8), 1, ConfigDeps) of
+             {_, V} when is_binary(V); is_list(V) ->
```
ferd: this clause obscures the one after since you kept the is_list(V) guard, and won't be converting the value of a list to a binary anymore. I'm not sure this is what you intended?

filmor: It wasn't, just a mistake when pushing. I'll fix it.

starbelly:
Hi, @filmor did you have a case where a version was not a binary? We get the version numbers from the projects rebar.lock file and those are stored in binary.

filmor:
This is the code path that takes the version specification from the `rebar.config`, if specified there. And the error was the other way round, the code only works with a `list` while `rebar3` handles binaries in the `{deps, [...]}.` just fine.

starbelly:
Ahh yeah, gotcha. So, later in the execution path we do call `binarify/1` on version... so I'm still not sure there was a bug, but I'm not sure there wasn't either... do you have a case?

filmor:
The problem is not the content of the version, it's the guard expression here. This commit shows a diff that was required: https://github.com/rusterlium/erlang-cargo/commit/9a2b12ebc5ac4395670cebbfd3defba6fbb69c71

The corresponding hex packages are here: https://hex.pm/packages/cargo
The diff between the pushed packages is this: https://diff.hex.pm/diff/cargo/0.1.0..0.1.1

You can see how the only (real) difference between these two is that I converted the dependency line from `{jsx, <<"~> 2.8">>}` to `{jsx, "~> 2.8"}` and this resulted in a corrected `hex_metadata.config`.

starbelly:
Gotcha. I wasn't aware that this binary for the version in rebar3 is actually allowed, thus it should be allowed here as well.

starbelly (approved):

tsloughter:
@filmor curious, what is the mix.exs for?

Issue #57, opened 2016-09-18
Make docs task configurable


yurrriq:
e.g. `lfe lodox` instead of `edoc`

tsloughter:
Good point.

yurrriq:
Turns out we're probably gonna use the EDoc format internally going forward so this will likely become a non-issue. Closing for now. Feel free to reopen.

yurrriq:
I don't need it immediately, but this would be cool still.

PR #180, opened 2020-08-17
[#177] publish does not include docs

### LINKED ISSUE #177: Publish does not build the docs
`rebar3 hex publish` publishes docs by default, but actually doesn't build the docs before. A separate run of `rebar3 hex docs` as to be done to actually publish the latest docs.

jfacorro:
Fixes #177

### DIFF
--- a/src/rebar3_hex.erl
+++ b/src/rebar3_hex.erl
@@ -1,6 +1,12 @@
 -module(rebar3_hex).
 
--export([init/1, gather_opts/2, get_required/2, task_args/1, repo_opt/0, help_opt/0]).
+-export([ init/1
+        , gather_opts/2
+        , get_required/2
+        , task_args/1
+        , repo_opt/0
+        , help_opt/0
+        ]).
 
 init(State) ->
     lists:foldl(fun provider_init/2, {ok, State}, [rebar3_hex_user,
@@ -17,13 +23,13 @@ init(State) ->
 provider_init(Module, {ok, State}) ->
     Module:init(State).
 
-gather_opts(Targets, State) -> 
+gather_opts(Targets, State) ->
     {Args, _} = rebar_state:command_parsed_args(State),
-    lists:foldl(fun(T, Acc) -> 
+    lists:foldl(fun(T, Acc) ->
                         case proplists:get_value(T, Args, undefined) of
-                            undefined -> 
+                            undefined ->
                                 Acc;
-                            V -> 
+                            V ->
                                 maps:put(T, V, Acc)
                         end
                 end, #{}, Targets).
--- a/src/rebar3_hex_client.erl
+++ b/src/rebar3_hex_client.erl
@@ -1,18 +1,19 @@
 -module(rebar3_hex_client).
 
 -export([ is_success/1
-         , key_add/3
-         , key_get/2
-         , key_delete/2
-         , key_delete_all/1
-         , key_list/1
-         , publish/3
-         , publish_docs/4
-         , delete_docs/3
-         , test_key/2
-         , member_of/1
-         , pretty_print_status/1
-         , pretty_print_errors/1]).
+        , key_add/3
+        , key_get/2
+        , key_delete/2
+        , key_delete_all/1
+        , key_list/1
+        , publish/3
+        , publish_docs/4
+        , delete_docs/3
+        , test_key/2
+        , member_of/1
+        , pretty_print_status/1
+        , pretty_print_errors/1
+        ]).
 
 -include("rebar3_hex.hrl").
 
@@ -29,7 +30,6 @@ key_get(HexConfig, <<KeyName/binary>>) ->
     response(Res);
 key_get(HexConfig, KeyName) ->
     key_get(HexConfig, to_binary(KeyName)).
-
 
 member_of(HexConfig) ->
     Res =  hex_api_organization:list(HexConfig),
--- a/src/rebar3_hex_config.erl
+++ b/src/rebar3_hex_config.erl
@@ -1,15 +1,15 @@
 -module(rebar3_hex_config).
 
--export([api_key_name/1
-         , api_key_name/2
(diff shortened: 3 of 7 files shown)


src/rebar3_hex_docs.erl
```diff
@@ -8,9 +8,10 @@
 -include("rebar3_hex.hrl").
 
 -define(PROVIDER, docs).
--define(DEPS, [{default, edoc}]).
+-define(DEPS, [{default, lock}]).
```
jfacorro, line 11: Removing the dependency on `edoc` addresses #57 in a way.

src/rebar3_hex_docs.erl
```diff
@@ -118,23 +142,32 @@ do_revert(App, State, Repo, Vsn) ->
...
+    AppOpts = rebar_app_info:opts(AppInfo),
+    EdocOpts = rebar_opts:get(AppOpts, edoc_opts, []),
+    AppDetails = rebar_app_info:app_details(AppInfo),
+    Dir = proplists:get_value(dir, EdocOpts, ?DEFAULT_DOC_DIR),
+    proplists:get_value(doc, AppDetails, Dir).
+
+-spec assert_doc_dir(string()) -> true.
+assert_doc_dir(DocDir) ->
+    filelib:is_dir(DocDir) orelse
+        rebar_api:abort( "Docs were not published since they "
```
jfacorro, line 162: Since the `publish` command calls the `rebar3_hex_docs:publish/3` function once it has already succeeded publishing the package itself, I though it would be better to use `rebar_api:abort/2` for showing the user the error and aborting the execution. This will only abort the docs publishing and not the package publishing.

src/rebar3_hex.erl
```diff
@@ -1,6 +1,12 @@
 -module(rebar3_hex).
```
jfacorro, line 1: Whitespace only changes.

src/rebar3_hex_client.erl
```diff
@@ -1,18 +1,19 @@
 -module(rebar3_hex_client).
```
jfacorro, line 1: Whitespace only changes.

src/rebar3_hex_config.erl
```diff
@@ -1,15 +1,15 @@
 -module(rebar3_hex_config).
```
jfacorro, line 1: Mostly whitespace only changes.

src/rebar3_hex_file.erl
```diff
@@ -1,8 +1,7 @@
 -module(rebar3_hex_file).
```
jfacorro, line 1: Whitespace only changes.

src/rebar3_hex_publish.erl
```diff
@@ -3,25 +3,28 @@
 %% @end
```
jfacorro, line 3: Whitespace only changes.

src/rebar3_hex_docs.erl
starbelly: Yeah I wonder if we actually want to bail on the publish 🤔 Like, if you're expecting it all in one go, then I could see where we might want to bail on the package publishing too, but I don't have strong feelings about it either.

starbelly: Yeah, I thought about this more and I really think we should strive for parity with hex (mix hex) here. Specifically, calling `rebar3 hex publish` assumes you want to publish both, we should check that docs exist before we attempt to publish anything, if they do not, bail and give the user a message.

We should also provide switches so inform publish to only publish the package or only publish the docs. `--package` and `--docs`. 

The message I would expect in the case that we see no docs and bail would be something like what you have + `If you really only want to publish the package without docs use the --package switch`

Thoughts?

jfacorro: I think that is a good idea, but I think the discussion around the details of implementing this new approach (i.e. `publish = pkg+docs` plus `--package` and `--docs` flags) deserves its own issue and separate PR. There might be many details that need to be discussed, for example, it would be nice to show the docs directory that will be used when listing the details on what is going to be published. 

Since this PR address the bug in the current behaviour discussed in #180 I would like to get it merged to avoid the unexpected results we get today.

tsloughter: To be clear, since I'm missing some context, we are running the `edoc` provider before publishing, right? If not we should have publish depending on or calling the edoc provider from rebar3 and then doing the publish.

If there aren't any docs even though it ran edoc then it should just continue on I would think, without erroring. Obviously will be an issue of someone overriding where edoc outputs shit, but whatever, can deal with that later.

jfacorro: The `publish` command currently also publishes the docs by calling the internal function in `rebar3_hex_docs:publish3`. It is in this situation that the published docs are empty.

The `docs` command depends on `{default, edoc}`. This PR removes that dependency to allow users to rely on other tools to generate documentation.

There is more context in the related issue #177.

tsloughter: Ah ok. Yea, I guess that function should call the edoc provider instead of having the provider use a dependency on it.

starbelly: Yeah, I'm cool with merging this now per what @jfacorro points out, i.e., it fixes a very confusing issue.

starbelly (approved):

jfacorro:
@starbelly @tsloughter Would it be possible to merge this? Or are there any changes that you think should be applied?

starbelly (changes requested):
I plan on merging this today, I was waiting to see what @tsloughter thought overall, but I think he's inundated with rebar3 itself. I'm changing this back to requested changes as I need to test the bug myself and QA this branch. Will have an update later this evening.

starbelly (approved):
@jfacorro okie dokie! Tested this, looks good to me... we'll do a new minor version to cover the change in behavior.

Issue #181, opened 2020-09-01
Safer default behavior for publishing. 


starbelly:
We recently merged a PR that alters the behavior when it comes to publishing packages and docs (see #177 and #180 ). 

As part of the conversation around that PR are some of my thoughts one where to go from here: 

 The current behavior of rebar3_hex is to attempt to publish both package and docs. We publish the package and *then* attempt to publish the docs. As part of #180 we will not attempt to publish docs if a docs dir is not found. 

 I think we should strive for parity with hex (mix hex) here. Specifically,  we should check that docs exist before we attempt to publish anything, if they do not, bail and give the user a message vs publishing the package and then trying to publish docs. This makes a publishing package and docs somewhat atomic. 

 To go with this we should provide switches for the publish command that allow a user to indicate explicitly that they only want to publish one or the other, such as `--package` and `--docs`.

tsloughter:
Does "not attempt to publish docs if docs dir is not found" refer to after having run `rebar3 edoc`? If that is the case, then that is fine. But if it checks for a `docs` dir before running `edoc` that is wrong.

But yea, strive for parity with mix hex.

starbelly:
@tsloughter yeah, wasn't very clear there and thoughts are bit muddy per the recent change.

Looking again, the operation doesn't seem to be exactly as I described. I looked at the code a few weeks ago, but it doesn't align with what I had thought of today : 

https://github.com/hexpm/hex/blob/771386942f4e5a928153040b22426916c5fa522c/lib/mix/tasks/hex.publish.ex#L182

So, they do try to publish the package first, then publish the docs, but there's a key difference in that you can somewhat guarantee the docs will get built, iirc `mix hex.publish` will error out of if you don't have ex_doc as dependency, and the chances of ex_doc failing is low, so really the main thing that could go wrong is in sending up the docs to the api. 

My take for `rebar3 hex publish` is by default : 

1. create the release 
2. create the docs
3. publish the release
4. publish the docs

Operation 3 doesn't happen unless 1 and 2 happen successfully. You can override this behavior with the switches mentioned in above though. Naturally, operation 4 doesn't happen if 3 fails, in which case we can tell the user to run `rebar3 hex publish --docs` or `rebar3 hex docs`  

IMO, this how mix hex should operate too, but that's another story and project :) 

I think the issue is dependent on what we decide on in #182.  Thoughts?

starbelly:
[...]

PR #185, opened 2020-10-17
add support for configuring docs provider

### LINKED ISSUE #57: Make docs task configurable

### LINKED ISSUE #182: Offer the user a way to configure a docs command or provider.
With the latest release of rebar3_hex (v6.10.0) we no longer assume you want to run edoc. This makes sense depending on your preferences, organization, and even BEAM language. 

However, most users expect a plugin such as rebar3_hex to handle this all nice and automagic so to speak. We should also strive for parity with mix hex in this regard. 

Perhaps something like :

Provider: 
```
{docs, edoc}.
```

Shell cmd: 
```
{docs, {cmd, "some_cmd_we_shell_out_to", ["-a", "-b", "--output_dir", "docs"]}}.
```

Just initial thoughts on config, probably a better format for it. Could also do `{hex_docs, ...}` to keep `docs` open or if `docs` is already reserved in rebar3.

Regardless of what one uses to generate docs, we should should check to see if `docs` dir has been made, etc.

starbelly:
This provides the minimal amount of support to fulfil #57 and support #181 . It can be extended to support edoc opts, maybe that should be done here in this PR, but I figured just go with the MVP. Other cases will arise and we can add support for them as needed (i.e., other providers, options, shell commands, etc.).

### DIFF
--- a/src/rebar3_hex_docs.erl
+++ b/src/rebar3_hex_docs.erl
@@ -98,6 +98,7 @@ handle_command(App, State, Repo) ->
     end.
 
 do_publish(App, State, Repo) ->
+    maybe_gen_docs(State),
     AppDir = rebar_app_info:dir(App),
     DocDir = resolve_doc_dir(App),
     assert_doc_dir(filename:join(AppDir, DocDir)),
@@ -155,6 +156,32 @@ resolve_doc_dir(AppInfo) ->
     AppDetails = rebar_app_info:app_details(AppInfo),
     Dir = proplists:get_value(dir, EdocOpts, ?DEFAULT_DOC_DIR),
     proplists:get_value(doc, AppDetails, Dir).
+
+maybe_gen_docs(State) -> 
+    case doc_opts(State) of
+        undefined -> 
+             rebar_api:error("No docs config found", []),
+            noop;
+        PrvName -> 
+            AllProviders = rebar_state:providers(State),
+             case providers:get_provider(PrvName, AllProviders) of
+                 not_found -> 
+                     rebar_api:error("No provider found for ~ts", [PrvName]),
+                     noop;
+                 Prv ->
+                     providers:do(Prv, State)
+             end
+    end.
+
+doc_opts(State) -> 
+    Opts = rebar_state:opts(State),
+    case proplists:get_value(doc, rebar_opts:get(Opts, hex), undefined) of
+        PrvName when is_atom(PrvName) -> PrvName;
+        undefined -> undefined;
+        _ -> 
+            %% Any other data type or structure is currently not supported. 
+            undefined
+    end.
 
 -spec assert_doc_dir(string()) -> true.
 assert_doc_dir(DocDir) ->
--- a/src/rebar3_hex_docs.erl
+++ b/src/rebar3_hex_docs.erl
@@ -192,7 +192,7 @@ maybe_post_process({shell, [{cmd, Cmd}, {args, Args}]}) ->
     rebar_utils:sh(Cmd2, [{use_stdout, true}, debug_and_abort_on_error]);
 maybe_post_process({shell, Cmd}) ->
     rebar_utils:sh(post_proc_cmd(Cmd), [{use_stdout, true}, debug_and_abort_on_error]);
-maybe_post_process(Eh) ->
+maybe_post_process(_) ->
     ok.
 
 post_proc_cmd(Cmd) ->
--- a/src/rebar3_hex_docs.erl
+++ b/src/rebar3_hex_docs.erl
@@ -160,17 +160,14 @@ resolve_doc_dir(AppInfo) ->
 maybe_gen_docs(State) ->
     case doc_opts(State) of
         {ok, {PrvName, Opts}} ->
-            AllProviders = rebar_state:providers(State),
-            case providers:get_provider(PrvName, AllProviders) of
+            case providers:get_provider(PrvName, rebar_state:providers(State)) of
                 not_found ->
-                    rebar_api:error("No provider found for ~ts", [PrvName]),
-                    noop;
+                    rebar_api:error("No provider found for ~ts", [PrvName]);
                 Prv ->
                     gen_docs(State, Prv, Opts)
             end;
         _ ->
-            rebar_api:error("No valid hex docs configuration found", []),
-            noop
+            rebar_api:error("No valid hex docs configuration found", [])
     end.
 
 gen_docs(State, Prv, Opts) ->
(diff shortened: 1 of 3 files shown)


src/rebar3_hex_docs.erl
```diff
@@ -156,6 +157,32 @@ resolve_doc_dir(AppInfo) ->
...
     proplists:get_value(doc, AppDetails, Dir).
 
+maybe_gen_docs(State) -> 
+    case doc_opts(State) of
+        undefined -> 
+             rebar_api:error("No docs config found", []),
+            noop;
+        PrvName -> 
+            AllProviders = rebar_state:providers(State),
+             case providers:get_provider(PrvName, AllProviders) of
```
starbelly: We may want to abstract the direct use of `providers` in rebar, maybe in rebar_api.

src/rebar3_hex_docs.erl
```diff
@@ -156,6 +157,32 @@ resolve_doc_dir(AppInfo) ->
...
+                 not_found -> 
+                     rebar_api:error("No provider found for ~ts", [PrvName]),
+                     noop;
+                 Prv ->
+                     providers:do(Prv, State)
+             end
+    end.
+
+doc_opts(State) -> 
+    Opts = rebar_state:opts(State),
```
starbelly: Maybe need to get this info from `App`vs `State` otherwise we could be operating on the wrong app 🤔

src/rebar3_hex_docs.erl
ferd: I don't think that this would be necessary. If we change from providers we break all plugins regardless since they all use it directly to instantiate themselves.

src/rebar3_hex_docs.erl
ferd: That's a fair concern. That being said if you're publishing to hex the assumption is that you're in an app-specific directory so in practice it shouldn't be a problem. If you're in an umbrella app you shouldn't be able to publish that easily right? Usually you `cd` into the app dir and do your thing there.

src/rebar3_hex_docs.erl
```diff
@@ -156,6 +157,32 @@ resolve_doc_dir(AppInfo) ->
...
+    case proplists:get_value(doc, rebar_opts:get(Opts, hex), undefined) of
+        PrvName when is_atom(PrvName) -> PrvName;
```
ferd: should wrap in `{ok, PrvName}` to avoid the case of `undefined` being a module that doesn't work :P

ferd (approved):

src/rebar3_hex_docs.erl
starbelly: Yeah, we actually removed the support for publishing umbrellas iirc too, so we're good 👍

src/rebar3_hex_docs.erl
```diff
@@ -156,6 +157,54 @@ resolve_doc_dir(AppInfo) ->
...
+                undefined ->
+                    ok;
+                PostOpts ->
+                    maybe_post_process(PostOpts)
+            end;
+        Err ->
+            ?PRV_ERROR({publish, Err})
+    end.
+
+maybe_post_process({shell, Cmd}) ->
```
starbelly: I'm quite sure dragons live here...

test/rebar3_hex_integration_SUITE.erl
```diff
@@ -119,22 +121,43 @@ docs_test(Config) ->
...
+docs_post_process_simple_test(Config) ->
+    P = #{app => "valid", mocks => [docs]},
+    {ok, #{rebar_state := State, repo := Repo}} = setup_state(P, Config),
+    Command = ["docs"],
+    RepoConfig =  [{repos,[Repo]}, {docs, {edoc, [{post_process, [{cmd, "echo eh?"}]}]}}],
+    {ok, DocState} = test_utils:mock_command(rebar3_hex_docs, Command, RepoConfig, State),
+    {ok, NewState} = rebar_prv_edoc:do(DocState),
+    ?assertMatch({ok, NewState}, rebar3_hex_docs:do(NewState)).
+
+docs_post_process_simple_error_test(Config) ->
```
starbelly: Le sigh, not sure about this test. It doesn't really tell us much other than "this doesn't blow up", I suppose that's valuable and inline with a lot of other tests in this suite.

test/rebar3_hex_integration_SUITE.erl
```diff
@@ -119,22 +123,59 @@ docs_test(Config) ->
...
     {ok, #{rebar_state := State, repo := Repo}} = setup_state(P, Config),
     Command = ["docs"],
-    {ok, DocState} = test_utils:mock_command(rebar3_hex_docs, Command, Repo, State),
+    RepoConfig =  [{repos,[Repo]}],
+    {ok, DocState} = test_utils:mock_command(rebar3_hex_docs, Command, RepoConfig, State),
     {ok, NewState} = rebar_prv_edoc:do(DocState),
     ?assertMatch({ok, NewState}, rebar3_hex_docs:do(NewState)).
 
+docs_post_process_test(Config) ->
+    P = #{app => "valid", mocks => [docs]},
```
starbelly: 🤔 This always fail with a rebar_abort exception, tested outside of test suite, works fine. Maybe `rebar_utils:sh/2` can't reliably be used in common tests?

test/rebar3_hex_integration_SUITE.erl
```diff
@@ -119,22 +123,59 @@ docs_test(Config) ->
...
+
+docs_post_process_error_test(Config) ->
+    P = #{app => "valid", mocks => [docs]},
+    {ok, #{rebar_state := State, repo := Repo}} = setup_state(P, Config),
+    Command = ["docs"],
+    RepoConfig =  [{repos,[Repo]}, {docs, {edoc, [{post_process, [{shell, "eh?"}]}]}}],
+    {ok, DocState} = test_utils:mock_command(rebar3_hex_docs, Command, RepoConfig, State),
+    ?assertThrow(rebar_abort, rebar3_hex_docs:do(DocState)).
+
+docs_post_process_with_args_test(Config) ->
```
starbelly: Ditto for this tests. Always fails, but works fine outside of CT.

starbelly:
I have reverted back to the original support, but with support for both `global` and `repo` specific config. The moduledoc will not be published ofc, but I think it helps review and will serve as a base template for updating the docs on rebar3.org. 

I think a re-review is in order at this point.

ferd (approved):

PR #186, opened 2020-10-18
Do not fail fast when calling hex_tarball:create/2

### LINKED ISSUE #179: Audit for fail fast cases
Fail fast is great! Except if your part of command line application. We need to audit the code for cases where we might crash and provide nothing useful to the end user.

starbelly:
- expect that hex_tarball:create/2 might fail
 - add format_error clause for when tarball size is too big, delegate to hex_tarball to format the error

Example output: 

```
Publishing truecoat 0.42.2 to hexpm
  Description: It gets installed at the factory
  Dependencies:

  Included files:
    CHANGELOG.md
    LICENSE
    README.md
    priv/VSCode-darwin-stable.zip
    rebar.config
    rebar.lock
    src/truecoat.app.src
    src/truecoat.erl
  Licenses: Apache 2.0
  Links:

  Build tools: rebar3
Proceed? ("Y")> y
Local Password:
===> package exceeds max uncompressed size 67.108864 MB
```


src/rebar3_hex_publish.erl
```diff
@@ -114,6 +114,8 @@ format_error(no_write_key) ->
     "No write key found for user. Be sure to authenticate first with:"
     ++ " rebar3 hex user auth";
 
+format_error({publish, {error, {tarball, too_big}}}) -> 
+    "Overall size of package exceeds max allowed size (10MB)";
```
starbelly: Message could probably be better. I think hex_core should return the max_size or make it available via a function (`hex_tarball:max_size/0`), but this works for the moment.

starbelly: Side note: The max size in this message is not correct. It's actually more like 8.3 megabytes for the tarball and 67 megabytes for the max uncompressed size.

starbelly:
Relates to : https://github.com/hexpm/hex_core/issues/100

src/rebar3_hex_publish.erl
ferd: might put it as `"Overall size of package exceeds max allowed size (8MB compressed, 67MB total)";` until we can get more accurate values. It won't error out if we're right above the 8mb size either but will clearly explain things if you go above. Whereas here you could get a 9mb package and a confusing error message.

tsloughter: Yea, that'd be better.

starbelly: So think I'm about to do a PR up to hex_core to return the size and add a format_error clause on the hex_core end, this way we don't have to worry about the values changing. Thus, I think that should go in, get hex_core to do a release, update this PR etc. Thoughts?

tsloughter: Yea, that'd be better.

starbelly: https://github.com/hexpm/hex_core/pull/103

starbelly: Second thought, let's just go with this for now. The PR to hex_core is in, but don't want to rush hex_core on a new release nor have to do a lot of QA per the bigger changes that just went into hex_core.

starbelly: Strike the above, they did a release.

starbelly: PR updated.

ferd (approved):

PR #189, opened 2020-10-27
Use crypto_one_time_aead on OTP 23 and onwards

starbelly:
- updated rebar3_hex_user to use crypto:crypto_one_time_aead/6 and
 crypto:crypto_one_time_aead/7 if OTP is equal to or greater than 23

 - added private function cipher/1 to determine the correct cipher to
 use with crypto:crypto_one_time_aead/6 and crypto:crypto_one_time_aead/7

 - standardize on upper case platform_define(s) in rebar.config

### DIFF
--- a/src/rebar3_hex_user.erl
+++ b/src/rebar3_hex_user.erl
@@ -241,14 +241,14 @@ generate_all_keys(Username, Password, LocalPassword, Repo, State) ->
     end.
 
 -ifdef(POST_OTP_22).
--spec encrypt_write_key(binary(), binary(), binary()) -> {binary(), binary(), binary(), {binary(), binary()}}.
+-spec encrypt_write_key(binary(), binary(), binary()) -> {binary(), {binary(), binary()}}.
 encrypt_write_key(Username, LocalPassword, WriteKey) ->
     AAD = Username,
     IV = crypto:strong_rand_bytes(16),
     Key =  pad(LocalPassword),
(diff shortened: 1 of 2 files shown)


paulo-ferraz-oliveira:
Great stuff. I approve these changes. Would it be possible to request for a tag post-merge? (projects with `warnings_as_errors` thank you)

starbelly:
@paulo-ferraz-oliveira Yes, will definitely do a release after this is merged to get this and a few other patch level changes out.

src/rebar3_hex_user.erl
```diff
@@ -240,25 +240,52 @@ generate_all_keys(Username, Password, LocalPassword, Repo, State) ->
...
         Result ->
             Result
     end.
+-endif.
+
+-ifdef(POST_OTP_22).
+cipher(Key) when size(Key) == 16  -> aes_128_gcm;
+cipher(Key) when size(Key) == 24  -> aes_192_gcm;
+cipher(Key) when size(Key) == 32  -> aes_256_gcm.
+-endif.
```
ferd: You want to use byte_size or bit_size, since size is not always clear (ie for a tuple it returns tuple_size)

starbelly: Yeah, initially just looked at what they were doing in crypto (https://github.com/erlang/otp/blob/238c8965d65aa7c7d2b93d82b411892d3f7bf2a8/lib/crypto/src/crypto.erl#L1529), but byte_size makes sense for us 👍 will adjust.

PR #200, opened 2021-03-04
Fix owner list package

paulo-ferraz-oliveira:
There's two fixes to this:
1. if a package doesn't exist we don't crash, but instead output `Error listing owners of package PACKAGE: Entity not found (404)` (where `PACKAGE` is the queried one),
2. if a package's owner doesn't have his email listed we don't crash[1]

---
[1] regarding this I don't know if 1. this is an Hex issue, that shouldn't return an empty e-mail field, an `httpc` issue (because the field is defined as `nil`) or a `rebar3_hex` issue, in which case the fix might be acceptable.

### DIFF
--- a/src/rebar3_hex_owner.erl
+++ b/src/rebar3_hex_owner.erl
@@ -168,7 +168,7 @@ remove(HexConfig, Package, UsernameOrEmail, State) ->
 list(HexConfig, Package, State) ->
     case hex_api_package_owner:list(HexConfig, Package) of
         {ok, {200, _Headers, List}} ->
-            Owners = [binary_to_list(maps:get(<<"email">>, Owner, <<"">>)) || Owner <- List],
+            Owners = [owner_email(Owner) || Owner <- List],
             OwnersString = rebar_string:join(Owners, "\n"),
             rebar3_hex_io:say("~s", [OwnersString]),
             {ok, State};
@@ -176,6 +176,12 @@ list(HexConfig, Package, State) ->
             ?PRV_ERROR({status, Status, Package});
         {error, Reason} ->
             ?PRV_ERROR({error, Package, Reason})
+    end.
+
+owner_email(Owner) ->
+    case maps:get(<<"email">>, Owner, <<"">>) of
+        Email when is_binary(Email) -> binary_to_list(Email);
+        _ -> "unspecified"
     end.
 
 verb_to_gerund(add) -> "adding";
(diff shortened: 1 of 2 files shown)


src/rebar3_hex_owner.erl
```diff
@@ -179,6 +178,12 @@ list(HexConfig, Package, State) ->
             ?PRV_ERROR({error, Package, Reason})
     end.
 
+owner_email(Owner) ->
+    case maps:get(<<"email">>, Owner, <<"">>) of
```
starbelly: I think the default here should be `undefined` otherwise you'll print out a blank new line in the end, thoughts?

src/rebar3_hex_owner.erl
```diff
@@ -179,6 +178,12 @@ list(HexConfig, Package, State) ->
...
+        Email when is_binary(Email) -> binary_to_list(Email);
+        _ -> "unspecified"
```
starbelly: Technically, we can opt to print the username here vs the email, what do you think?

src/rebar3_hex_owner.erl
paulo-ferraz-oliveira: I guess we could change it. I just moved the code, so didn't think about prior reasoning behind it.

src/rebar3_hex_owner.erl
paulo-ferraz-oliveira: Does that field always exist? If so, we could output `Name (Email)` always.

src/rebar3_hex_owner.erl
paulo-ferraz-oliveira: Hm... now that I think about it. I think that "" is what's actually being output as `nil`, but this is just an assumption. We could also default to `nil` which is what causes the issue to start with 😄

src/rebar3_hex_owner.erl
starbelly: I like that idea! One I like even better is printing it as a table, but I think that's another PR.

paulo-ferraz-oliveira:
Most recent commit might just address both your comments at once. Lemme know.

starbelly:
@paulo-ferraz-oliveira I believe they do. Build is failing for some reason (not circle-ci)

paulo-ferraz-oliveira:
Lemme check this.

starbelly (approved):

paulo-ferraz-oliveira:
Thanks much.

PR #210, opened 2021-03-17
Add --without-docs option for publish task

starbelly:
- Add a --without-docs option for the publish task which defaults to
   false. If switched on, will only publish a package regardless of configuration.

### DIFF
--- a/src/rebar3_hex_publish.erl
+++ b/src/rebar3_hex_publish.erl
@@ -49,7 +49,8 @@ init(State) ->
                                          {yes, $y, "yes", {boolean, false}, help(yes)},
                                          {replace, undefined, "replace", {boolean, false}, help(replace)},
                                          {package, $p, "package", string, help(package)},
-                                         {revert, undefined, "revert", string, help(revert)}]}]),
+                                         {revert, undefined, "revert", string, help(revert)},
+                                         {without_docs, undefined, "without-docs", {boolean, false}, help(without_docs)}]}]),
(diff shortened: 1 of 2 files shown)


ferd (approved):
Logic is sound.

Would it make sense to have it as `--with-docs=true|false` and defaulting to `true`, so that instead of writing `--without-docs=true` you just write `--with-docs=false` (or maybe even just `--docs`)

starbelly:
> Logic is sound.
> 
> Would it make sense to have it as `--with-docs=true|false` and defaulting to `true`, so that instead of writing `--without-docs=true` you just write `--with-docs=false` (or maybe even just `--docs`)

We could go that route, but you don't have to specify `--without-docs=true`, just `--without-docs`. The idea being you'll only ever use that switch if you want it on. Does that make sense? 

Edit:

Example output: 

```shell
===> Analyzing applications...
===> Compiling rebar3_hex
===> Analyzing applications...
===> Compiling rebar3_hex
Publishing truecoat 0.42.42 to hexpm
  Description: It gets installed at the factory
  Dependencies:

  Included files:
    CHANGELOG.md
    LICENSE
    README.md
    rebar.config
    rebar.lock
    src/truecoat.app.src
    src/truecoat.erl
  Licenses: Apache 2.0
  Links:

  Build tools: rebar3
Proceed? ("Y")> y
Local Password:
===> Published truecoat 0.42.42
===> --without-docs is enabled : will not publish docs
```

ferd:
Yeah that's fine too

PR #212, opened 2021-03-17
Adjust docs dir check assertion to check for index.html

starbelly:
- We were checking only for the presence of doc dir but not to see if  it at least contained an index.html, this commit refactors the private function on the docs task to check for that and aborts if missing.


test/rebar3_hex_integration_SUITE_data/test_apps/doc_dir_missing_index/.gitignore
```diff
@@ -0,0 +1,19 @@
+.rebar3
```
starbelly: This is lazy, but I was having trouble getting the path to the test app and wanted to get this in quick for re-publishing 6.11.0

starbelly:
@ferd last one for the night

src/rebar3_hex_docs.erl
```diff
@@ -240,13 +240,17 @@ gen_docs(State, Prv) ->
 
 -spec assert_doc_dir(string()) -> true.
 assert_doc_dir(DocDir) ->
-    filelib:is_dir(DocDir) orelse
-        rebar_api:abort( "Docs were not published since they "
+    case rebar_utils:find_files(DocDir, "index.html") of
+        [_] ->
+            true;
+         _ ->
```
ferd: `rebar_utils:find_files/2` works on a regex and is recursive in its search by default. You could therefore find `["index.html", "subdir/index.html", "subdir2/other-index.html"]` and fail the check because you only look for single-item lists.

starbelly: ahh! Not good. I switched to this function because older versions of OTP don't have `filelib:find_file/2`. Can make a wrapper, but also #213

test/rebar3_hex_integration_SUITE_data/test_apps/doc_dir_missing_index/.gitignore
starbelly: Switched to filelib:is_file/1, this should be solid now I believe.

src/rebar3_hex_docs.erl
starbelly: Switched to filelib:is_file/1, this should be solid now I believe.

ferd (approved):

PR #219, opened 2021-04-03
Switch to setup-beam github action

starbelly:
- changes ci.yml to use setup-beam for all tests, except 19 which is still an ongoing issue with setup-beam

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        otp: ['20.3.8.9', '21.3.8.9', '22.3.4.9', '23.3.1']
+        otp: ['20', '21', '22', '23']
         rebar3: ['3.14.1', '3.14.2', '3.14.3']
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       matrix:
         otp: ['20', '21', '22', '23']
-        rebar3: ['3.14.1', '3.14.2', '3.14.3', '3.14.3']
+        rebar3: ['3.14.1', '3.14.2', '3.14.3', '3.14.4']
     steps:
       - uses: actions/checkout@v2
       - uses: erlef/setup-beam@v1.7.0
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,23 +1,46 @@
 ---
-name: build
+name: test
 on:
   push:
     branches:
       - master
   pull_request:
-    branches:
-      - master
+    types: [opened, synchronize]
+
 jobs:
-  ci:
-    name: >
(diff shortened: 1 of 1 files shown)


starbelly:
We will wait until monday to merge so we can set the version in our config to `@v1`

.github/workflows/ci.yml
```diff
@@ -1,23 +1,44 @@
...
-    container:
-      image: erlang:${{matrix.otp_vsn}}
+  test:
+    name: Erlang/OTP ${{matrix.otp}} / rebar3 ${{matrix.rebar3}}
+    runs-on: ubuntu-latest
     strategy:
       matrix:
-        otp_vsn: [19.3, 20.3, 21.3, 22.3, 23.2]
-        os: [ubuntu-latest]
+        otp: ['20.3.8.9', '21.3.8.9', '22.3.4.9', '23.3.1']
```
paulo-ferraz-oliveira: Why not use '20', instead of '20.3.8.9'? The action will pick the latest for you 😄

starbelly: It probably should yeah, I copied what was done in verl and that was just testing a good bit for the most part, but yeah agree to use majors only.

starbelly:
We should go ahead and merge this, we can simply do another PR to change to `@v1` when its done.

.github/workflows/ci.yml
```diff
@@ -1,23 +1,44 @@
...
-      image: erlang:${{matrix.otp_vsn}}
+  test:
+    name: Erlang/OTP ${{matrix.otp}} / rebar3 ${{matrix.rebar3}}
+    runs-on: ubuntu-latest
     strategy:
       matrix:
-        otp_vsn: [19.3, 20.3, 21.3, 22.3, 23.2]
-        os: [ubuntu-latest]
+        otp: ['20', '21', '22', '23']
+        rebar3: ['3.14.1', '3.14.2', '3.14.3']
```
paulo-ferraz-oliveira: FWIW, **latest** is 3.14.4.

paulo-ferraz-oliveira (approved):

.github/workflows/ci.yml
starbelly: Yeah, need to add.

.github/workflows/ci.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     strategy:
       matrix:
         otp: ['20', '21', '22', '23']
-        rebar3: ['3.14.1', '3.14.2', '3.14.3']
+        rebar3: ['3.14.1', '3.14.2', '3.14.3', '3.14.3']
```
paulo-ferraz-oliveira:
```suggestion
        rebar3: ['3.14.1', '3.14.2', '3.14.3', '3.14.4']
```

paulo-ferraz-oliveira: 😄

starbelly: One of those days 😁

paulo-ferraz-oliveira: I kinda guessed...

ericmj (approved):

PR #227, opened 2021-05-14
Increase consumer confidence

paulo-ferraz-oliveira:


paulo-ferraz-oliveira:
I'm also up'ing `rebar3`, since the version of vendored `hex_core` it's shipping with is not 24.0 compatible.

paulo-ferraz-oliveira:
That did it! Squashing.

ferd (approved):

PR #229, opened 2021-05-20
CI OTP 24

paulo-ferraz-oliveira:
Can't actually remember why I wanted/needed to introduce `dialyzer` here, but it was somehow on my TODO list.

The proposed changes only work with a yet unreleased `hex_core` which also caused `dialyzer` warnings, but is a good preview of what's to come.

Here it is:

- [x] wait for https://github.com/hexpm/hex_core/pull/121 to be merged and published
- [x] wait for https://github.com/jelly-beam/verl/pull/21 to be merged and published

Opening as draft for the above mentioned reasons.

Commenting on the stuff that might not be so obvious (or on which I have doubts).


rebar.config
```diff
@@ -10,7 +10,7 @@
 
 {project_plugins, [covertool, {rebar3_hank, "~> 0.3.0"}]}.
 
-{deps, [{hex_core, "0.7.1"}, {verl, "1.0.2"}]}.
+{deps, [{hex_core, "0.8.0"}, {verl, "1.1.0"}]}.
```
paulo-ferraz-oliveira: Ah, yeah, `verl` also needs to be bumped; I'll add that to the description.

rebar.config
```diff
@@ -23,6 +23,14 @@
     ]}
 ]}.
 
+{dialyzer, [
+    {warnings, [
+        error_handling,
```
paulo-ferraz-oliveira, line 28: Let me know if we want `unhandled_returns`. `unknown` is usually not friendly to `rebar3` plugins.

src/rebar3_hex_docs.erl
```diff
@@ -146,7 +146,7 @@ do_revert(App, State, Repo, Vsn) ->
     Name = rebar_utils:to_list(rebar_app_info:name(App)),
     PkgName = rebar_utils:to_list(proplists:get_value(pkg_name, AppDetails, Name)),
     case rebar3_hex_client:delete_docs(Config, rebar_utils:to_binary(PkgName), rebar_utils:to_binary(Vsn)) of
-        {modified, _Body} ->
+        {ok, _Body} ->
```
paulo-ferraz-oliveira: `Body`'ll be, most likely `nil`. This was confirmed between me and @starbelly.

src/rebar3_hex_error.erl
```diff
@@ -21,5 +21,5 @@ format_error({Cmd, missing_required_params}) ->
     io_lib:format("Required parameters for the ~ts command have not been supplied.", [Cmd]);
 
 format_error(Reason) ->
-    rebar_api:debug("Unknown error : ~ts", [Reason]),
+    rebar_api:debug("Unknown error : ~ps", [Reason]),
```
paulo-ferraz-oliveira, line 24: Would throw when the error is e.g. a tuple. (it was not `dialyzer`, but testing on an invalid repo.)

src/rebar3_hex_publish.erl
```diff
@@ -419,17 +419,17 @@ is_empty_prop(K, PropList) ->
           false
     end.
 
-is_unstable_dep({_, {pkg, Pkg, Ver, _, _}, _}, Acc) -> 
+is_unstable_dep({_, {pkg, Pkg, Ver, _, _}, _}, Acc) ->
```
paulo-ferraz-oliveira, line 422: Unwarranted white space deletions. Let me know if I should revert or create a separate commit.

src/rebar3_hex_user.erl
```diff
@@ -208,7 +208,7 @@ generate_all_keys(Username, Password, LocalPassword, Repo, State) ->
 
     %% write key
     WriteKeyName = api_key_name(),
-    WritePermissions = [#{<<"domain">> => <<"api">>}],
+    WritePermissions = [#{domain => api}],
```
paulo-ferraz-oliveira, line 215: This comes from `hex_core`'s API. I'm pretty sure it works in both forms, but, as-is, it's converted later, via `term_to_binary` eventually (if I correctly followed the 🐰 hole).

src/rebar3_hex_publish.erl
starbelly: It's fine with me.

src/rebar3_hex_docs.erl
starbelly: This may get fixed in #232. Because it introduces a hex_http adapter which uses hex_http behavior and because I didn't want to turn off dialyzer, etc. I opted to fix it up there. We'll see what the morning brings.

paulo-ferraz-oliveira: Sure, I'll rebase and retry no worries. Thanks.

paulo-ferraz-oliveira:
Following https://github.com/erlef/rebar3_hex/pull/232, this has been rebased on top of `master`. Initial conditions, https://github.com/erlef/rebar3_hex/pull/229#issue-648967795, still apply, though.

paulo-ferraz-oliveira:
hexpm/hex_core#121 was merged and tagged, so I pushed a remade `rebar.lock`.
After jelly-beam/verl#21, CI should be all 🟢. I've been testing this locally with `_checkouts`.

paulo-ferraz-oliveira:
jelly-beam/verl#21 was merged and tagged, so I pushed a remade `rebar.lock`. If CI's Okay that does it for my change 😄

paulo-ferraz-oliveira:
I also fixed coverage stuff. It was hidden by the fact it didn't show an error.
But look at README and see it's not there.

starbelly (approved):

PR #232, opened 2021-05-25
Add custom hex httpc adapter

starbelly:
Per https://ferd.ca/you-ve-got-to-upgrade-rebar3.html we must provide our own http adapter so we can ensure ssl opts are set. Note that we can provide the default hex core httpc adapter ssl options, however having a custom adapter will provide us greater flexibility long term. This especially true in the case of custom repositories for which the edge cases around ssl are unknown.

- Added rebar3_hex_httpc_adapter copied from rebar3.
- Changed rebar3_hex_config to use rebar3_hex_httpc_adapter instead of the default hex_core http adapter
- Changed test/support/test_utils.erl to work with rebar3 version 3.16
- Changed .github/workflows/ci.yml to only test against 3.16.1
- Fixed a bug in rebar3_hex_docs where our call to rebar3_hex_client:delete_docs/3 had wrong expectations
- Suppressed some dialyzer warnings that should be fixed up in a subsequent PR

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,8 +13,8 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        otp: ['20', '21', '22', '23', '24']
-        rebar3: ['3.14.2', '3.14.3', '3.14.4', '3.15.0', '3.15.1']
+        otp: ['22', '23', '24']
+        rebar3: ['3.16.1']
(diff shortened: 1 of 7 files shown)


ferd (approved):
I notice the only supporting rebar3 3.16.1 but that's only for tests due to internal API changes and should cause no practical concerns otherwise.

Fix is good for security.

PR #235, opened 2021-06-13
Add V7 publish options

starbelly:
This commit constitutes the breaking interface changes for the publish command for V7.

Options aside, one of the biggest changes is removing umbrella related functionality. Specifically, instead of looking
for all possible apps and iterating over them and optionally publishing them, we just get the current app and publish it.

In addition, rebar3_hex:command_state/1 has been added, it's not extensively used yet. We will expand on this function,
what it returns exactly, and proliferate its usage in the publish command and all other commands in subsequent
commits.

 - add package task on publish command
 - add docs task support on publish command
 - remove --without-docs option on publish command
 - remove --package option that goes with --revert on publish command
   - This was only ever needed per umbrella support
 - Remove umbrella support (i.e., we only publish the current app vs all possible apps)
 - add command_state/1 on rebar3_hex


starbelly:
Note : Aside from command state usage and refactoring / clean up that needs to be done, we need to add --docs-dir option. My gut told me to do this in another PR to keep this small enough. 

Edit:

To note rebar3_hex:get_required/2 , rebar3_hex:task_args/1, and rebar3_hex:gather_opts/2 will all be going away in subsequent PRs.

src/rebar3_hex.erl
```diff
@@ -51,6 +52,45 @@ task_args(State) ->
             {Task, proplists:delete(task, Opts)}
     end.
 
+command_state(State) ->
```
starbelly: Calling this `command_state` isn't exactly true, but isn't exactly wrong either. It really shouldn't be a big deal, but wanted thoughts on this so we can have consistent language going forward. If we look at what the user sees :   `rebar3 help [<namespace>] <task>`, however if I say `rebar3 help hex publish package`, then `package` gets recognized as the task.  I suppose we can  say `publish` is the default task, while `package` is a sub-task. 

So here I'm trying to make a clear distinction between the two saying `publish` is the command which may have a task argument. 

Alternatively, we can call this `task_state/1` and when we encounter a `task` tuple we could call that `command` or `sub_task` or something along those lines. What ever is going to be the least confusing for anyone who might hack on this and also in regards to documentation. 

Thoughts?

ferd: I don't fully understand what you mean by `rebar3 help hex publish package` with `package` being the task; I sort of assume the task is publishing and that `package` is the type / an option / a parameter. I figure more specifically it specifies the handler or provider to run, which is more of an implementation detail.

I guess in general that's because `publish` is different from the other commands, where you use `mix hex.owner add|transfer|remove|...`, `mix hex.repo add|set|remove|list`, `mix hex.user auth|deauth|key <op>|...` where the general structure is `mix hex.<entity> <operation>` whereas publishing is clashing by being `mix hex.<operation> <entity>`. The `search` and `retire` commands similarly clash there.

if we line up with the original implementation, we can't necessarily hope to be clearer than they are. We might do a broader category as you said of `<task> <subtask>`, which might be as clear as we can get. Any arbitrary terms of this type should work, like `<category> <provider>`, `<group> <command>`, `<meta> <command>`, etc. As long as you carry in that meaning of "one container for the specific" you should be good to go.

starbelly: Ahh please ignore, I saw `{task, package}` ending up somewhere but that's because of something silly I was doing. I'm going to rename this to `task_state/1`

src/rebar3_hex.erl
```diff
@@ -51,6 +57,47 @@ task_args(State) ->
             {Task, proplists:delete(task, Opts)}
     end.
 
+-spec task_state(rebar_state:t()) -> {ok, task()} | {error, term()}.
+task_state(State) ->
```
starbelly, line 61: I don't particularly like the name of this function, but it can be changed.

src/rebar3_hex.erl
```diff
@@ -1,13 +1,19 @@
...
+        , get_args/1
         , gather_opts/2
         , get_required/2
         , task_args/1
+        , task_state/1
         , repo_opt/0
         , help_opt/0
         ]).
 
+-type task() :: #{args := map(), repo := map(), state := rebar_state:t()}.
```
starbelly, line 13: I opted not to call this `task_state` or `state` as in `rebar3_hex:state/0`. I also opted not to name the key `state`, `rebar_state`. This is all internal, no one should be using this as a lib, and thus it's just implementation details. I think it would be confusing to have too many things called `*state*`. Likewise, `state` in a rebar3 plugin is synonymous with rebar3 state to me. Open to suggestion if there's strong feelings. Of course, like a lot of things we can always change it again while busting up the interface and refactoring further.

ferd (approved):

PR #236, opened 2021-06-16
Warn for existence of _checkouts

### LINKED ISSUE #68: Publishing should probably ignore _checkouts
Currently, when a dependency is put into `_checkouts` for development, `rebar3_hex` will silently remove that from the dependencies when trying to publish. It should probably prevent that or ignore the directory for this run.

paulo-ferraz-oliveira:
Closes #68.

~I'm not sure a warning is enough, but also think `rebar3` can't stop on `:init` (at least, I know not how to), so I chose this.~

~I'm proposing we simply ignore it, as suggested by @starbelly below.~

Re-reading the linked issue, this doesn't seem right. I'm again going for the warning approach.

### DIFF
--- a/src/rebar3_hex_publish.erl
+++ b/src/rebar3_hex_publish.erl
@@ -218,11 +218,16 @@ publish(App, Name, Version,  Deps, [], AppDetails, HexConfig, State) ->
     rebar3_hex_io:say("  Build tools: ~ts", [format_build_tools(BuildTools)]),
     maybe_say_coc(HexConfig),
     {Args, _} = rebar_state:command_parsed_args(State),
+    MaybeWithWarnings =
+        case has_checkouts_for(AppDir) of
+            {false, _} -> "";
+            {true, _} -> " (with warnings)"
+        end,
     case proplists:get_bool(yes, Args) of
         true ->
             publish_package_and_docs(Name, Version, Metadata, PackageFiles, HexConfig, App, State);
         false ->
-            case rebar3_hex_io:ask("Proceed?", boolean, "Y") of
+            case rebar3_hex_io:ask("Proceed" ++ MaybeWithWarnings ++ "?", boolean, "Y") of
                 true ->
                     publish_package_and_docs(Name, Version, Metadata, PackageFiles, HexConfig, App, State);
                 _ ->
@@ -326,6 +331,8 @@ to_map(List) when is_list(List) ->
     maps:from_list(List).
 
 include_files(Name, AppDir, AppDetails) ->
+    _ = maybe_print_checkouts_warnings(AppDir),
+
     AppSrc = {application, to_atom(Name), AppDetails},
     FilePaths = proplists:get_value(files, AppDetails, ?DEFAULT_FILES),
     IncludeFilePaths = proplists:get_value(include_files, AppDetails, []),
@@ -347,6 +354,14 @@ include_files(Name, AppDir, AppDetails) ->
     AppSrcBinary = rebar_utils:to_binary(lists:flatten(io_lib:format("~tp.\n", [AppSrc]))),
     lists:keystore(AppFileSrc, 1, WithIncludes, {AppFileSrc, AppSrcBinary}).
 
+maybe_print_checkouts_warnings(AppDir) ->
+    {HasCheckouts, Checkouts} = has_checkouts_for(AppDir),
+    HasCheckouts andalso
+        rebar_log:log(warn, "~p directory found; this might interfere with publishing", [Checkouts]).
+
+has_checkouts_for(AppDir) ->
+    Checkouts = filename:join(AppDir, "_checkouts"),
+    {filelib:is_dir(Checkouts), Checkouts}.
 
 is_valid_app({_App, _Name, _Version, _AppDetails, _Deps} = A) ->
     F = fun(K, Acc) ->


src/rebar3_hex.erl
```diff
@@ -15,6 +15,8 @@
 -export_type([task/0]).
 
 init(State) ->
+    filelib:is_dir("_checkouts") andalso
+        rebar_log:log(warn, "_checkouts directory found; this might interfere with publishing", []),
```
ferd: The provider is initialized at rebar3 load time. This will warn on every single execution of rebar3, including `rebar3 help`.

ferd (changes requested):
This warning mode is too aggressive.

src/rebar3_hex.erl
paulo-ferraz-oliveira: Oh, hahaha. Where should it reside better, then?

paulo-ferraz-oliveira: Should it be per `do`?

starbelly: This should happen in `rebar3_hex_publish`, but we might want to make it an opt in. I would need more information on the case though. It seems like we're wanting to warn if a `_checkouts` dir simply exist in the project you're publishing for. Is the concern that you might accidentally publish the `_checkouts` dir as an artifact? 

Edit : 

Maybe a good way to tackle this is make it part of `known_exclude_file/2` in `rebar3_hex_publish`. I can't think of why someone would ever want to publish `_checkouts` as part of their project, but if there actually is a case we can make it so it has to be very explicit. Basically, make it part of our known files to exclude, don't warn if we see it, but if an author *really* wants to put it in, they can add it to included files list explicitly.

paulo-ferraz-oliveira: > This should happen in rebar3_hex_publish

I thought about this, but then there's `cut` too, right? So I tried to make it "global" so as to not have to list all the options and forgot to add the warning to some new one that might come later. But I might have gone too far on that thinking...

paulo-ferraz-oliveira: > we might want to make it an opt in

For the warning or the exit? What would the default be, in that case? I think it's reasonable to always get the warning, and only potentially publish...

paulo-ferraz-oliveira: > Is the concern that you might accidentally publish the _checkouts dir as an artifact?

This is probably better expressed in #68.

paulo-ferraz-oliveira: > Maybe a good way to tackle this is make it part of known_exclude_file/2 in rebar3_hex_publish.

I went with Fred's initial suggestion on warning, but can experiment with yours too. Lemme do some stuff here.

paulo-ferraz-oliveira: (rethinking this, I'd go with the warning, since the problem seems to be with `_checkouts`-deps not being locked, which is a `rebar3` feature, and not publishing `_checkouts` as @starbelly thought it might be)

starbelly: As FYI `cut` is going away in v7.

paulo-ferraz-oliveira: Sure. Fun stuff.

paulo-ferraz-oliveira: This is also especially nice since cutting a patch was tagging versions with a `v` prefix, which I don't use 😄

src/rebar3_hex_publish.erl
```diff
@@ -157,22 +157,28 @@ format_error(Reason) ->
 %% ===================================================================
 
 publish(App, HexConfig, State) ->
-    Name = rebar_app_info:name(App),
+    Publish = not(filelib:is_dir("_checkouts")),
```
paulo-ferraz-oliveira: Best seen with "Hide whitespace changes".

src/rebar3_hex_publish.erl
```diff
@@ -157,22 +157,28 @@ format_error(Reason) ->
...
-            ?PRV_ERROR(Errors)
+            case is_valid_app({App, Name, ResolvedVersion, AppDetails, Deps}) of
+                ok ->
+                    publish(App, Name, ResolvedVersion, TopLevel,
+                            Excluded, AppDetails, HexConfig, State);
+                {error, Errors} ->
+                    ?PRV_ERROR(Errors)
+            end;
+        _ ->
+            ?PRV_ERROR({error, checkouts})
```
paulo-ferraz-oliveira: I'm aware this is very generic, but it's also the only no-publish condition, at the moment...

paulo-ferraz-oliveira:
This errors out with
```
✗ rebar3 hex publish
===> Analyzing applications...
===> Compiling rebar3_hex
===> _checkouts directory found; this might interfere with publishing
```

We can always add an option to not error out and assume the publisher knows what it's doing.

starbelly:
> This errors out with
> 
> ```
> ✗ rebar3 hex publish
> ===> Analyzing applications...
> ===> Compiling rebar3_hex
> ===> Verifying dependencies...
> ===> _checkouts directory found; this might interfere with publishing
> ```
> 
> We can always add an option to not error out and assume the publisher knows what it's doing.

So I think I'd prefer a warning myself. Let me give you a case 🧐 ... the case of publishing rebar3_hex itself. We intentionally put a `_checkouts` dir in rebar3_hex when publishing a new version as a form of QA if you will. Sometimes because we have to, can't remember off hand that "have to" condition, but yeah still a valid case. Thus it's best if we just ignore the dir and warn about it IMO. Ignoring it is simple ofc.

paulo-ferraz-oliveira:
> So I think I'd prefer a warning myself. Let me give you a case 🧐 ... the case of publishing rebar3_hex itself. We intentionally put a `_checkouts` dir in rebar3_hex when publishing a new version as a form of QA if you will. Sometimes because we have to, can't remember off hand that "have to" condition, but yeah still a valid case. Thus it's best if we just ignore the dir and warn about it IMO. Ignoring it is simple ofc.

I believe it's ignored already, isn't it? Otherwise, how are you doing it in your use case?

As for the warning, sure, but @tsloughter just pointed out, earlier, that it might get hidden if a release is huge and involves tons of files.

**Edit**: moved back to a non-blocking/stopping warning.

starbelly:
> As for the warning, sure, but @tsloughter just pointed out, earlier, that it might get hidden if a release is huge and involves tons of files.

Yeah, that makes sense, which is why we should filter out `_checkouts` dir when we expand paths and build up dirs and files for the tarball that gets created. 

Edit:

Forgot to note, right now we don't explicitly ignore `_checkouts`, rather we just don't put it in the default includes (see line 22 in rebar3_hex_publish). 

But Tristan raises a case that hasn't been addressed, but could be by explicitly excluding any  `_checkouts` dir. So take the case of someone having a `_checkouts` dir in `src` (i.e., `src/_checkouts` (contrived example), that would end up in the tarball. We don't want that, and if someone really would want it, we should make it so they can explicitly opt in to include it. 

So, with how you're warning is right now, it's just case the simplest case possible, which is fine for a start and this PR IMO. 

Now, how does `_checkouts` work when we publish rebar3_hex? 

1. cd into rebar3_hex dir
2. mkdir _checkouts
3. copy the contents of rebar3_hex into _checkout/rebar3_hex or do a new git clone
4. rebar3 hex publish

rebar3 will load rebar3_hex vs using the plugin that's in deps or global plugins. I probably was a bit confusing when I talked about that, so figured a better explanation of that case was due.

paulo-ferraz-oliveira:
> Yeah, that makes sense, which is why we should filter out _checkouts dir when we expand paths and build up dirs and files for the tarball that gets created.

In that sense, a warning is not required, right? Unless it is to tell you what you're publishing might not work as you expect it to.

paulo-ferraz-oliveira:
> we should make it so they can explicitly opt in to include it.

Does an option for this already exist?

paulo-ferraz-oliveira:
> which is fine for a start and this PR IMO.

I don't mind adding more, but I need explicit orientation. You want a warning + excluding _checkouts?

starbelly:
> I don't mind adding more, but I need explicit orientation. You want a warning + excluding _checkouts?

Yup, warn + exclude, which resolves #68

paulo-ferraz-oliveira:
Hm... if I just add `_checkouts$` to the filter it'll probably filter more than `<cwd>/_checkouts`. Any proper way you want that or are you Ok with just `_checkouts$`?

starbelly:
> Hm... if I just add `_checkouts$` to the filter it'll probably filter more than `<cwd>/_checkouts`. Any proper way you want that or are you Ok with just `_checkouts$`?

Yeah, that would be the idea, catch all the instances of `_checkouts`. Further, warn about each instance found. I think a warning is prudent as an author might actually have a valid case where they do not want to exclude `_checkouts` dir. Giving them a warning would at least give them the chance to back out of the operation and modify their config to include `_checkouts` and we will have had made a best effort at sending them down the "right" path.

paulo-ferraz-oliveira:
Sure, but the warning would be printed potentially before the file/folder listing so might get lost, at least visually, right?

paulo-ferraz-oliveira:
So this is what it's doing now:
1. it warns about the existence of `_checkouts` (per app, even in umbrella),
2. it changes the "Proceed?" message to "Proceed (with warnings)?" where warranted.

src/rebar3_hex_publish.erl
```diff
@@ -326,6 +331,8 @@ to_map(List) when is_list(List) ->
     maps:from_list(List).
 
 include_files(Name, AppDir, AppDetails) ->
+    _ = maybe_print__checkouts_warnings(AppDir),
```
starbelly: Nit: I suggest to call this `maybe_print_checkouts_warnings` vs `maybe_print__checkouts_warnings`

src/rebar3_hex_publish.erl
```diff
@@ -347,6 +354,14 @@ include_files(Name, AppDir, AppDetails) ->
     AppSrcBinary = rebar_utils:to_binary(lists:flatten(io_lib:format("~tp.\n", [AppSrc]))),
     lists:keystore(AppFileSrc, 1, WithIncludes, {AppFileSrc, AppSrcBinary}).
 
+maybe_print__checkouts_warnings(AppDir) ->
+    {HasCheckouts, Checkouts} = has__checkouts_for(AppDir),
+    HasCheckouts andalso
+        rebar_log:log(warn, "~p directory found; this might interfere with publishing", [Checkouts]).
+
+has__checkouts_for(AppDir) ->
```
starbelly: Nit : same, remove the `_` before `checkouts`. I realize the dir is called `_checkouts`, but not sure it makes sense as part of a function name. 🤷

starbelly (changes requested):

src/rebar3_hex_publish.erl
paulo-ferraz-oliveira: I'm Ok with this, though its' just as you say: cosmetics.

src/rebar3_hex_publish.erl
paulo-ferraz-oliveira: Yeah, I wanted to be correct, but if it pleases you otherwise I'll change it.

paulo-ferraz-oliveira: "Fixed", rebased and force-pushed.

src/rebar3_hex_publish.erl
paulo-ferraz-oliveira: "Fixed", rebased and force-pushed.

starbelly (approved):

ferd (approved):

Issue #221, opened 2021-04-09
v7.0.0


starbelly:
# Version 7

This issue enumerates everything that needs to be changed in order to be aligned with mix/hex interface and behavior that would be considered a breaking change. Updating docs on rebar3.org is implicit for each item and required. 

## Breaking changes 

- [x] add `docs ` arg support on publish (i.e., `rebar3 hex publish docs`)
- [x] add `package` arg on publish task  (i.e., `rebar3 hex publish package`)
- [x] remove `--without-docs` option on publish 
- [x] remove `--package` option that goes with `--revert`
- [x] add `--app` switch on `publish` task for specifying a single app to in an umbrella for all publish sub tasks. 

- [x] add `--docs-dir` option on publish task (unique to rebar3_hex)
- [x]  support `--dry-run` for publishing in general
- [x] Merge existing docs code into `publish` task and remove umbrella related code
- [x] Merge existing `revert` task code into publish task
- [x] Update `cut` task to work with `publish` task changes. 

- [x] re-work repo task to be aligned with mix hex organization
- [x] move `key` task code to `user` task and remove `key` task
- [x] add unretire support on `rebar3_hex_retire` 
- [x] change `message` argument to a switch in `rebar3_hex_retire` 
- [x] - add a `build`  task ( see https://hexdocs.pm/hex/Mix.Tasks.Hex.Build.html#content)

## Tasks that should be completed before release
- [x] Add docs
- [x]  QA

## Internal changes that do not have to happen at any particular stage

The following items are to be completed prior to additive changes to ease contributions by others. These need to be further defined (properly defined) in their own respective issues or draft PR(s). 

- vendor in hex_core
- refactor all existing task modules to follow a common pattern (rebar3_hex_key is a prototype)
- refactor existing functions into smaller functions where applicable, we have a lot of functions that are simply too large. 
- expose or move all functions (which also means making new functions) that do not have side effects in modules / into new modules and test them directly 
 - per the above do as little in rebar3_hex_integration_SUITE as possible, it should be limited to what can not be tested any other way
- contributing / developer docs , some what internal, but should be made available on rebar3_hex repo

## Additive changes - post ver 7.0.0 release

The following items are added here for enumeration purposes. An issue is to be created for each item. These are prioritized and should be completed in order. Items that are nice-to-have are just that, they are not crucial or critical in any form. 

- add a config task (existing issue https://github.com/tsloughter/rebar3_hex/issues/208)
[...]

PR #237, opened 2021-06-16
Remove cut task

starbelly:
As part of V7 the cut command is removed in this commit. We can bring it back later or perhaps better it becomes its own plugin.

- remove rebar3_hex_cut.erl
- remove rebar3_hex_cut from rebar3_hex:init/1
- remove reference to cut to in rebar3_hex_publish


paulo-ferraz-oliveira (approved):

tsloughter:
Uh, what? Why? Is this functionality provided by another command now? If not, please add this back.

paulo-ferraz-oliveira:
Ref: https://github.com/erlef/rebar3_hex/issues/221#issuecomment-835420930

PR #241, opened 2021-06-19
Restore release (umbrella) publish support

starbelly:
Obsoletes #240 

This is the bare minimum to restore support for publishing from the top level of a release. That is to say, it essentially
puts things back how they were vs including refactors that will happen in this module.

- Restore support for publishing multiple apps in a release
- Add --app switch which acts as the old --package switch did (required for reverting a specific app)
- Set the current_app in rebar state via rebar3_hex:task_state/1 so that rebar_state:current_app/1
  may be utilized when not in a release.
- Set a multi_project k/v (boolean) in rebar3_hex:task_state/1

### DIFF
--- a/src/rebar3_hex.erl
+++ b/src/rebar3_hex.erl
@@ -10,7 +10,7 @@
         , help_opt/0
         ]).
 
--type task() :: #{args := map(), repo := map(), state := rebar_state:t(), multi_project := boolean()}.
+-type task() :: #{args := map(), repo := map(), state := rebar_state:t(), multi_app := boolean()}.
 
 -export_type([task/0]).
 
--- a/src/rebar3_hex.erl
+++ b/src/rebar3_hex.erl
@@ -62,7 +62,13 @@ task_state(State) ->
      case rebar3_hex_config:repo(State) of
          {ok, Repo} -> 
              Opts = get_args(State),
-             {ok, #{args => Opts, repo => Repo, state => State}};
+             case rebar_state:project_apps(State) of 
+                 [App] ->
+                     State1 = rebar_state:current_app(State, App),
+                    {ok, #{args => Opts, repo => Repo, state => State1, multi_project => false}};
+                 [_|_] -> 
+                    {ok, #{args => Opts, repo => Repo, state => State, multi_project => true}}
+             end;
          Err -> 
             Err
      end.
--- a/src/rebar3_hex.erl
+++ b/src/rebar3_hex.erl
@@ -10,7 +10,7 @@
         , help_opt/0
         ]).
 
--type task() :: #{args := map(), repo := map(), state := rebar_state:t()}.
+-type task() :: #{args := map(), repo := map(), state := rebar_state:t(), multi_project := boolean()}.
 
 -export_type([task/0]).
 
--- a/src/rebar3_hex.erl
+++ b/src/rebar3_hex.erl
@@ -65,9 +65,9 @@ task_state(State) ->
              case rebar_state:project_apps(State) of 
                  [App] ->
                      State1 = rebar_state:current_app(State, App),
-                    {ok, #{args => Opts, repo => Repo, state => State1, multi_project => false}};
+                    {ok, #{args => Opts, repo => Repo, state => State1, multi_app => false}};
                  [_|_] -> 
-                    {ok, #{args => Opts, repo => Repo, state => State, multi_project => true}}
+                    {ok, #{args => Opts, repo => Repo, state => State, multi_app => true}}
              end;
          Err -> 
             Err
--- a/src/rebar3_hex_publish.erl
+++ b/src/rebar3_hex_publish.erl
@@ -563,7 +563,7 @@ to_list(X)
 
 help(app) ->
     "Specifies the app to use with the publish command, currently only utilized for publish and revert operations"
-    "Note the app switch and value only has to be provided if you are publishing within a release (umbrella).";
+    "Note that the app switch and value only have to be provided if you are publishing within a release (umbrella).";
 help(revert) ->
     "Revert given version, if the last version is reverted the package is removed";
 help(replace) ->
--- a/src/rebar3_hex_publish.erl
+++ b/src/rebar3_hex_publish.erl
@@ -34,10 +34,6 @@
                      , has_unstable_deps
                      ]).
 
--ifdef(TEST).
--include_lib("eunit/include/eunit.hrl").
--endif.
-
 %% ===================================================================
 %% Public API
 %% ===================================================================
@@ -55,6 +51,7 @@ init(State) ->
                                  {desc, support()},
                                  {opts, [rebar3_hex:repo_opt(),
                                          {yes, $y, "yes", {boolean, false}, help(yes)},
+                                         {app, $a, "app", string, help(app)},
                                          {replace, undefined, "replace", {boolean, false}, help(replace)},
                                          {revert, undefined, "revert", string, help(revert)}]}]),
     State1 = rebar_state:add_provider(State, Provider),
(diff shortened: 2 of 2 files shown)


src/rebar3_hex.erl
```diff
@@ -62,7 +62,13 @@ task_state(State) ->
      case rebar3_hex_config:repo(State) of
          {ok, Repo} -> 
              Opts = get_args(State),
-             {ok, #{args => Opts, repo => Repo, state => State}};
+             case rebar_state:project_apps(State) of 
+                 [App] ->
+                     State1 = rebar_state:current_app(State, App),
```
starbelly, line 67: This is called `multi_project` vs say `is_release` because a release might have one sub-app to start with, thus we don't know if we're in a release or not, further would not be necessary to make that distinction in regards to deciding whether to give the user the option of publishing multiple apps or simply going with the one app found.

src/rebar3_hex_publish.erl
```diff
@@ -34,10 +34,6 @@
                      , has_unstable_deps
                      ]).
 
--ifdef(TEST).
```
starbelly, line 37: Old stuff, not needed.

src/rebar3_hex_publish.erl
```diff
@@ -69,21 +66,45 @@ do(State) ->
...
-handle_task(#{args := #{revert := Vsn}, repo := Repo, state := State}) ->
+handle_task(#{args := #{revert := Vsn}, multi_project := false} = Task) ->
+    #{repo := Repo, state := State} = Task,
     App = rebar_state:current_app(State),
     Name = rebar_app_info:name(App),
     ok = rebar3_hex_revert:revert(binarify(Name), binarify(Vsn), Repo, State),
     {ok, State};
 
-handle_task(#{repo := Repo, state := State}) ->
+handle_task(#{args := #{revert := Vsn, app := AppName}, multi_project := true} = Task) ->
```
starbelly, line 80: This is in a sub-optimal state still, but I wanted to unblock @paulo-ferraz-oliveira and get this in real quick vs refactoring right now.

src/rebar3_hex_publish.erl
```diff
@@ -69,21 +66,45 @@ do(State) ->
...
+    #{repo := Repo, state := State} = Task,
+    ok = rebar3_hex_revert:revert(binarify(AppName), binarify(Vsn), Repo, State),
+    {ok, State};
+
+handle_task(#{args := #{revert := _}, multi_project := true}) ->
+    {error, "--app required when reverting in a release with multiple projects"};
+
+%% Publish
```
starbelly, line 88: This is in a sub-optimal state still, but I wanted to unblock @paulo-ferraz-oliveira and get this in real quick vs refactoring right now.

src/rebar3_hex_publish.erl
```diff
@@ -538,6 +559,11 @@ to_list(X)
   when erlang:is_list(X) ->
     X.
 
+
+
+help(app) ->
+    "Specifies the app to use with the publish command, currently only utilized for publish and revert operations"
```
starbelly, line 565: `--app` isn't used by publish _yet_ . I'll be doing a PR tomorrow to make it so.

src/rebar3_hex_publish.erl
paulo-ferraz-oliveira: > I wanted to unblock @paulo-ferraz-oliveira
 
You mean because `publish` (atm) is broken?

src/rebar3_hex_publish.erl
```diff
@@ -69,21 +66,45 @@ do(State) ->
...
     {ok, State};
 
-handle_task(#{repo := Repo, state := State}) ->
+handle_task(#{args := #{revert := Vsn, app := AppName}, multi_project := true} = Task) ->
+    #{repo := Repo, state := State} = Task,
+    ok = rebar3_hex_revert:revert(binarify(AppName), binarify(Vsn), Repo, State),
+    {ok, State};
+
+handle_task(#{args := #{revert := _}, multi_project := true}) ->
+    {error, "--app required when reverting in a release with multiple projects"};
```
paulo-ferraz-oliveira:
```suggestion
    {error, "--app required when reverting a release with multiple projects"};
```

src/rebar3_hex.erl
tsloughter: Yes, it should not say anything about `release`. It is unrelated to anything release related.

I think `multi_app` would be better. You are in a single rebar3 project which may have multiple apps, but not multiple projects.

src/rebar3_hex_publish.erl
```diff
@@ -538,6 +559,11 @@ to_list(X)
   when erlang:is_list(X) ->
     X.
 
+
+
+help(app) ->
+    "Specifies the app to use with the publish command, currently only utilized for publish and revert operations"
+    "Note the app switch and value only has to be provided if you are publishing within a release (umbrella).";
```
paulo-ferraz-oliveira:
```suggestion
    "Note that the app switch and value only have to be provided if you are publishing within a release (umbrella).";
```

paulo-ferraz-oliveira: Should we never refer to `release`? I keep getting confused with this vs. a `relx` release. Umbrella is Okay, though, not ambiguous (to me).

paulo-ferraz-oliveira (approved):
Approved, pending some minor comments.

src/rebar3_hex_publish.erl
```diff
@@ -69,21 +66,45 @@ do(State) ->
...
+handle_task(#{repo := Repo, state := State, multi_project := false}) -> 
+    case maps:get(write_key, Repo, maps:get(api_key, Repo, undefined)) of
+            undefined ->
+                ?PRV_ERROR(no_write_key);
+            _ ->
+                App = rebar_state:current_app(State),
+                publish(App, Repo, State)
+        end;
+
+handle_task(#{repo := Repo, state := State, multi_project := true}) ->
```
tsloughter: This could be a single clause function if you create a function `select_apps` which when `multi_app` is false simply returns the current app in a list `[App]` and if `multi_app` is `true` then do the `select_app`, then run the `publish` over the list.

src/rebar3_hex.erl
paulo-ferraz-oliveira: @tsloughter, I think the confusion comes from the fact that `rebar3 new` has `release` and `umbrella` for basically the same time (probably even the same thing).

tsloughter: Yea, I dislike that. If it was possible to treat `rebar.config` as writable I'd 100% want to remove the `relx` section from `rebar.config` when generated for `rebar3 new umbrella` and would make `rebar3 new release` simply add a `relx` section to the current project's `rebar.config` and create the `config/` files.

I so often have a single app project (so with no `apps/` dir) that builds a release, but it is manual to go from a `rebar3 new app` to a rebar3 project that can build a release.

So yea, agree, probably should never have put the release stuff in `umbrella` template. I think there is a major documentation page missing that can go in detail into each of these.

src/rebar3_hex_publish.erl
tsloughter: Right, in `rebar3_hex` there should probably be no use of the term `release`.

src/rebar3_hex.erl
starbelly: > Yes, it should not say anything about `release`. It is unrelated to anything release related.
> 
> I think `multi_app` would be better. You are in a single rebar3 project which may have multiple apps, but not multiple projects.

Yup, like `multi_app` better 👍 ... and it's shorter 😁

src/rebar3_hex_publish.erl
starbelly: @paulo-ferraz-oliveira correct

src/rebar3_hex_publish.erl
starbelly: I think what I was shooting for was making the behavior obvious and up front, maybe that's a bit of overkill / obsessive passion though. TBH I'm still not sure exactly what that's going to look like, I'd have a better idea after doing some refactoring (today). If the results of refactoring show that it's better to have a single function clause, we can go with that, does that sound good?

tsloughter (approved):
I'd still refactor `handle_task` in publish module to be a single clause function and not rely on setting `current_app` -- just treat a single element list as the same as multi-element list of apps. But that isn't a reason to block this.

starbelly:
> I'd still refactor `handle_task` in publish module to be a single clause function and not rely on setting `current_app` -- just treat a single element list as the same as multi-element list of apps. But that isn't a reason to block this.

May end up doing just that, but let's see in my next PR if I can refactor such that the paths are clear, what's more if something goes wrong it's easily identifiable as belonging to a specific a path. It may be that I see after refactoring, this isn't worth it, it doesn't add any additional clarity.

PR #242, opened 2021-06-19
Change api domain value to binary in rebar3_hex_user

starbelly:
The domain value for write permissions should be a binary vs an atom.

Since upgrading to hex_core 0.8.2 the domain value must be a binary instead of an atom. I say hex_core 0.8.2 as that's the only thing that has changed. It's possible something changed on hexpm side as well, regardless, it should not be an atom 😁 

Edit:

It's not hex_core, hex_core still has the value defined as an atom, but it doesn't get converted, and hexpm wants a binary, not an atom. May be a bug in hex_core.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -16,7 +16,7 @@
     {test, [
         {extra_src_dirs, ["test/support"]},
         {overrides, [{override, rebar3,[{deps, [{erlware_commons, "1.3.1"}]}]}]},
-        {deps, [{hex_core, "0.7.1"},
+        {deps, [{hex_core, "0.8.2"},
                 {erlware_commons, "1.5.0"}, {elli, "3.3.0"},
                 {jsone, "1.5.3"}, {meck, "0.9.0"}]},
         {erl_opts, [nowarn_export_all]}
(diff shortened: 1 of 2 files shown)


starbelly:
@wojtekmach Is there a bug in hex_core either in the type spec or in protobufs? If hex_core is supposed to handle the conversion to a binary, then it's a bug bug, if it's not then it's a type spec bug. 

Edit:

Reference point : https://github.com/hexpm/hex_core/blob/master/src/hex_api_key.erl#L20

wojtekmach:
hex_core should standardise around ~strings~ binaries, see https://github.com/hexpm/hex_core/issues/71, so this change is definitely in the right direction.

starbelly:
> hex_core should standardise around ~strings~ binaries, see [hexpm/hex_core#71](https://github.com/hexpm/hex_core/issues/71), so this change is definitely in the right direction.

Right on, we can just tell dialyzer to ignore the this call until a new release of hex_core is done

paulo-ferraz-oliveira:
I'm not keen on all the `nowarn` stuff, but I guess you could create an issue to review this later, lest real warnings are hidden too.

paulo-ferraz-oliveira (approved):

starbelly:
Yeah, I'll make an issue. I'm also following the issue on hex_core.

PR #243, opened 2021-06-20
Fix crash for "no version for docs"

paulo-ferraz-oliveira:
This was crashing: I'm not sure whether that's the message you want, though, in `rebar3_hex_docs`.

Bonus fix was removing an extra `s` from a log message (in `rebar3_hex_error`).

Also, should `rebar3_hex_error`'s `format_error(Reason)` return a string so as to not crash? Or is the crash intended, here?


starbelly (approved):

starbelly:
> Also, should rebar3_hex_error's format_error(Reason) return a string so as to not crash? Or is the crash intended, here?

Forgot to note: We should never crash since this is a user facing application if you will. So, this change is right on the money and inline with many upcoming changes 👍

PR #244, opened 2021-06-20
Change api domain value to binary in rebar3_hex_user (#242)

starbelly:
The domain value for write permissions should be a binary vs an atom.

 -  Ignore some functions temporarily and bump hex_core to 0.8.2 in test


starbelly:
Cherry picked fix from master to 6.11 branch so we can release 6.11.5

paulo-ferraz-oliveira (approved):

PR #247, opened 2021-06-25
Refactor rebar3_hex_publish and support --app on publish

starbelly:
This commit is mostly a refactor of rebar3_hex_publish. The following rationale is provided for the refactoring:

## Refactor rationale 

 - Avoid going down paths and branches when we don't have to
   - If we only publish docs as an example, there's no need to validate the app, build up a package, etc.
 - Raising provider errors vs passing provider errors around
   - This is for all intents and purposes a command line application, as such it doesn't need
     to be treated like a general purpose library.
   - More importantly, by raising errors we can  ensure our error makes it back up to the caller (rebar3)
     and we get the error in our format_error/1 callback.
   - The above is important because we never want to crash in such a user facing application simply because
     we failed to handle an error case or properly propogate it up the stack ourselves.
   - Finally per raising instead of returning errors we end up with leaner code..
 - Splitting up big functions into smaller functions so that we may rely less on "integration" tests and
   can focus more on unit tests (e.g, tests with less mecking, more stubbing, and no fake servers).
 - To establish a pattern going forward for all other tasks (both existing and forth coming).

## Changes

 - Added ?RAISE_ERROR macro, raises a ?PROVIDER_ERROR
 - Added assert_valid_app/2 which raises when the app is not deemed valid
 - Move app validation up the stack closer to the entry point
 - Refactored publish/* into publish_package/3 and publish_docs/3
 - Added build_package/3
 - Added create_tarball/2
 - Changed task_state/1 to put project_apps in task() and to not set the current_app

## TODO: 

 - [x] Update rebar3_hex_cut to work with refactored functions
 - [x] Fix dialyzer warnings
 - [x] QA against a local hex instance


src/rebar3_hex_app.erl
```diff
@@ -0,0 +1,205 @@
...
+    case rebar3_hex_version:parse(Ver) of
+        {ok, #{pre := Pre}} when Pre =/= [] ->
+            [{Pkg, Ver} | Acc];
+        _ ->
+            Acc
+    end;
+is_unstable_dep(_, Acc) ->
+    Acc.
+
+-spec vcs_vsn(rebar_state:t(), rebar_app_info:t()) -> string().
```
starbelly, line 147: This function comes from `rebar_utils`.  It is deprecated and was only kept in rebar3 for rebar3_hex, just went ahead and moved it on in so that rebar3 may be able to remove it now. Note that this function doesn't live in the new rebar3_hex_version module as it's coupled with apps in a rebar3 project. It's also not pure, but can't win them all :)

src/rebar3_hex_app.erl
```diff
@@ -0,0 +1,205 @@
...
+    deps := list()
+}.
+
+-type validation_errors() :: #{
+    warnings => list(),
+    errors => list()
+}.
+
+-spec find([rebar_app_info:t()], string()) -> {ok, rebar_app_info:t()} | {error, app_not_found}.
+find(Apps, Name) ->
```
starbelly, line 30: I believe `rebar_app_utils` has a similar function, we could just rip this out and use it. No strong feelings.

src/rebar3_hex_app.erl
```diff
@@ -0,0 +1,205 @@
...
+    end,
+    case lists:filter(Pred, Apps) of
+        [App] ->
+            {ok, App};
+        _ ->
+            {error, app_not_found}
+    end.
+
+-spec get_deps(list()) -> {ok, list()} | {error, {non_hex_deps, list()}}.
+get_deps(Deps) ->
```
starbelly, line 42: This was moved here as it is can be said it's a pure function. One of the goals in this refactor was to start moving out pure functions into appropriate modules for better testing (i.e., not having to rely on so called integration tests so much, which currently only really tell us if something blew up or not).

src/rebar3_hex_app.erl
```diff
@@ -0,0 +1,205 @@
...
+
+lock_to_dep({A, {pkg, N, V, _, _}, 0}, Acc) ->
+    [{N, [{<<"app">>, A}, {<<"optional">>, false}, {<<"requirement">>, V}]} | Acc];
+lock_to_dep({A, {pkg, N, V, _}, 0}, Acc) ->
+    [{N, [{<<"app">>, A}, {<<"optional">>, false}, {<<"requirement">>, V}]} | Acc];
+lock_to_dep(_, Acc) ->
+    Acc.
+
+-spec validate(app_data()) -> ok | {error, validation_errors()}.
+validate(AppData) ->
```
starbelly, line 65: Ditto, pure function.

src/rebar3_hex_cut.erl
```diff
@@ -24,116 +24,134 @@ init(State) ->
...
         _ ->
             Spec = rebar3_hex_file:update_app_src(App, NewVersion),
             NewAppSrcFile = io_lib:format("~tp.\n", [Spec]),
             ok = rebar_file_utils:write_file_if_contents_differ(AppSrcFile, NewAppSrcFile),
             ask_commit_and_push(NewVersion),
-            case rebar3_hex_publish:publish(rebar_app_info:original_vsn(App, NewVersion), HexConfig, State) of
-                {ok, _State} ->
+            case try_publish(State, Repo, rebar_app_info:original_vsn(App, NewVersion), Args) of
+                {ok, _} ->
+                    maybe_create_and_push_tag(NewVersion),
```
starbelly, line 97: This one is new. Before we were not offering to create and push and a tag for this case... maybe we shouldn't but I thought it made sense.

src/rebar3_hex_cut.erl
```diff
@@ -142,36 +160,40 @@ get_increment(Version) ->
     case rebar3_hex_io:ask("[1-4] ", number) of
         4 ->
             rebar3_hex_io:ask("New Version ", string);
-        Type ->
-            int_to_bump(Type)
+        TypeInt ->
+            case int_to_increment(TypeInt) of
+                error ->
+                    rebar_api:error("Invalid number given, try again~n", []),
```
starbelly, line 166: New : If the users enters in a number that isn't 1-4, ask them again.

starbelly:
This is open for review now. The actual set of changes is small if you don't include deletion of stubs. One thing this PR lacks is testing for umbrellas, but we didn't have this already. We do need this, but it might require re-working some of the integration test helpers a good bit, and I didn't want this PR to grow anymore. 

More refactoring and splitting out pure functions into other modules should be a part of that, this should happen in a subsequent PR. More tests need to be added as well of course, but that should also be done as another PR IMO. 

I'd like to get this one through and then focus on the rest of the breaking changes in other tasks (and the refactoring of those modules) at this point.

src/rebar3_hex_error.erl
```diff
@@ -20,6 +20,10 @@ format_error({Cmd, unsupported_params}) ->
 format_error({Cmd, missing_required_params}) ->
     io_lib:format("Required parameters for the ~ts command have not been supplied.", [Cmd]);
 
+format_error({_Cmd, {error, {failed_connect, [{to_address,_},
```
starbelly, line 23: Should have had this for a while now.

src/rebar3_hex_publish.erl
```diff
@@ -174,38 +135,161 @@ format_error(Reason) ->
...
-    Name = rebar_app_info:name(App),
 
-    Version = rebar_app_info:original_vsn(App),
-    ResolvedVersion = rebar_utils:vcs_vsn(App, Version, State),
-    {application, _, AppDetails} = rebar3_hex_file:update_app_src(App, ResolvedVersion),
+%% ===================================================================
+%% Publish package only operations
+%% ===================================================================
+handle_task(#{args := #{task := package, app := undefined}, multi_app := true}) ->
+    ?RAISE({app_switch_required, "--app required when publishing with the package argument in a umbrella"});
```
starbelly, line 142: Rationale : If you're only wanting to publish a package or docs (not both) then you probably only want to do it on a single app basis, but would defer to @tsloughter

src/rebar3_hex_publish.erl
```diff
@@ -174,38 +135,161 @@ format_error(Reason) ->
...
 
+handle_task(#{args := #{task := docs}, apps := [App]} = Task) ->
+    #{args := Args, repo := Repo, state := State} = Task,
+    publish_docs(State, Repo, App, Args),
+    {ok, State};
 
-    Deps = rebar_state:get(State, {locks, default}, []),
-    {TopLevel, Excluded} = gather_deps(Deps),
+handle_task(#{args := #{task := docs, app := AppName}, apps := Apps} = Task) ->
+    case rebar3_hex_app:find(Apps, AppName) of
```
starbelly, line 172: Rationale : If you specified an app to publish or revert, let's go ahead and check if it actually exists.

src/rebar3_hex_version.erl
```diff
@@ -0,0 +1,30 @@
+%%% ===================================================================
+%%% Version utils
+%%% ===================================================================
```
starbelly, line 3: Not unlike rebar3_hex_app, this is just a module that will hold pure functions related to version parses, incrementing, and formatting.

src/rebar3_hex_publish.erl
```diff
@@ -225,88 +309,80 @@ publish(App, Name, Version,  Deps, [], AppDetails, HexConfig, State) ->
...
+                [] -> 
+                    {ok, State};
+                Errs -> 
+                    ?RAISE({validation_errors, Errs})
+            end
+    end.
 
 hex_opts(Opts) ->
-    lists:filter(fun({K, _}) -> is_hex_opt(K) end, Opts).
+    lists:filter(fun({K, _}) -> is_hex_opt(K) end, maps:to_list(Opts)).
```
starbelly, line 371: Changed because we now pass around a map of arguments vs a proplist.

paulo-ferraz-oliveira:
> Added ?RAISE_ERROR macro, raises a ?PROVIDER_ERROR

I think you mean `?RAISE`.

> Refactored publish/* into publish_package/3 and publish_docs/3

I think you mean `/4` on both instances.

> Added create_tarball/2

I think you mean `/1`.

(not trying to nitpick, just want to make sure the changelog is proper, if you ever have one)

paulo-ferraz-oliveira:
> Refactored publish/* into publish_package/3 and publish_docs/3

You mean `/4` on both instances, right?

paulo-ferraz-oliveira:
Did none of this imply changes to the README? Or will this come at a later date?

paulo-ferraz-oliveira (approved):

starbelly:
> > Added ?RAISE_ERROR macro, raises a ?PROVIDER_ERROR
> 
> I think you mean `?RAISE`.

Correct, will amend. 
> 
> > Refactored publish/* into publish_package/3 and publish_docs/3
> 
> I think you mean `/4` on both instances.

Correct, will amend. 
> 
> > Added create_tarball/2
> 
> I think you mean `/1`.

Correct, will amend. 
> 
> (not trying to nitpick, just want to make sure the changelog is proper, if you ever have one)

No worries, and yes, I have been thinking we need to start keeping a changelog. There's at least two ways to do that. Systematically (requires everyone to do commits in the same format) or we just cobble it together and prepend on each release, but yeah fully agree none the less.

starbelly:
> Did none of this imply changes to the README? Or will this come at a later date?

Later. There's still a good number of PRs that need to happen first, but it shall be updated. We'll also have to update the docs. That brings up another issue, which I will make an issue for.

paulo-ferraz-oliveira:
> No worries, and yes, I have been thinking we need to start keeping a changelog. There's at least two ways to do that. Systematically (requires everyone to do commits in the same format) or we just cobble it together and prepend on each release, but yeah fully agree none the less.

I've seen a bunch of ways this is done, in the past, in other projects.

1. a CHANGELOG.md maintained by hand,
2. a note on the GitHub releases page, also by hand,
3. a CHANGELOG.md maintained more or less automatically,
4. commits that follow a standard (very very easy to get wrong, unless automatically checked upon push - and am sure this would make some uncomfortable to contribute).

I prefer 3. If the PR titles are proper (I wouldn't go as deep as commits, since mostly we're going to squash them to 1 per pull request - right?), it should be fairly easy to generate one. And it doesn't even need a lot of discipline (unless the releaser wants to have as little trouble as possible), since you can always generate it, tweak the titles - if not readable -, then generate again. (to be discussed in another forum, perhaps)

starbelly:
@tsloughter ping for review

tsloughter (approved):

PR #255, opened 2021-07-23
Prevent compile-time deprecation warning while preparing for the future

paulo-ferraz-oliveira:
Mind you that `OTP_RELEASE` was introduced in Erlang 21, so if you want a broader maintenance scope I may have to tweak the pull request.


starbelly:
@paulo-ferraz-oliveira Yeah, might as well use a `platform_define` in rebar.config. `POST_OTP_22` already exists in that form (i.e., `-ifdef(POST_OTP_22)`).

paulo-ferraz-oliveira:
Rebased on top of `master`.

starbelly (approved):

PR #257, opened 2021-09-05
Improve local password collection

### LINKED ISSUE #190: Warn user when local password is greater than 32 bytes
The cryptographic cipher we use for encrypting hex keys has an upper bounds of key size of 32 bytes, yet we don't prompt the user to enter in something less than or equal to that if they enter something greater than said upper bounds. This leads to a crash. 

 Instead, if the user enters a local password that is > 32 bytes we should re-prompt them to re-enter a shorter local password.

### LINKED ISSUE #256: rebar3 hex crashes with local passwords over 32 characters in length.
Specifically, the functions `encrypt_write_key` and `decrypt_write_key` call `pad` ([here](https://github.com/erlef/rebar3_hex/blob/master/src/rebar3_hex_user.erl#L250) and [here](https://github.com/erlef/rebar3_hex/blob/master/src/rebar3_hex_user.erl#L269)), but pad uses a non-exhaustive case over the length of the input string.

This crashes all tasks that call either of these functions:
```shell
$ rebar3 hex user auth
[… output truncated]
Generating all keys...
===> Uncaught error in rebar_core. Run with DIAGNOSTIC=1 to see stacktrace or consult rebar3.crashdump
```

A simple workaround is of course to use only **local** passwords up to 32 characters in size (passwords for hex.pm are not affected by this problem), but the nondescript error and stacktrace make this limitation not obvious. 

The stacktrace from the crash, mainly so that other people hitting this problem have an easier time finding the workaround:

```erlang
Error: {case_clause,33}
[{rebar3_hex_user,pad,1,
                  [{file,"./_build/default/plugins/rebar3_hex/src/rebar3_hex_user.erl"},
                   {line,196}]},
 {rebar3_hex_user,encrypt_write_key,3,
                  [{file,"./_build/default/plugins/rebar3_hex/src/rebar3_hex_user.erl"},
                   {line,248}]},
 {rebar3_hex_user,generate_all_keys,5,
                  [{file,"./_build/default/plugins/rebar3_hex/src/rebar3_hex_user.erl"},
                   {line,217}]},
 {rebar_core,do,2,
             [{file,"/tmp/rebar3-4716517424079074236/rebar3/src/rebar3-3.16.1/src/rebar_core.erl"},
              {line,155}]},
 {rebar_prv_do,do_task,5,
               [{file,"/tmp/rebar3-4716517424079074236/rebar3/src/rebar3-3.16.1/src/rebar_prv_do.erl"},
                {line,87}]},
 {rebar_core,do,2,
             [{file,"/tmp/rebar3-4716517424079074236/rebar3/src/rebar3-3.16.1/src/rebar_core.erl"},
              {line,155}]},
 {rebar3,run_aux,2,
         [{file,"/tmp/rebar3-4716517424079074236/rebar3/src/rebar3-3.16.1/src/rebar3.erl"},
          {line,182}]},
 {rebar3,main,1,
         [{file,"/tmp/rebar3-4716517424079074236/rebar3/src/rebar3-3.16.1/src/rebar3.erl"},
          {line,66}]}]

```

starbelly:
Closes #256 and #190 

 - Resolves an issue where by if we accept a local password over 32 bytes we'll crash at key generation
  time. Instead, we now warn if the password is over 32 chars and prompt the user to try again. On the
  third attempt we throw a provider error.
 - To go along with the above, we now will warn the user if confirmation of a local password
   does not match the original entered local password vs immediately throwing, instead the user has 3 attempts
   to correctly confirm.

### DIFF
--- a/src/rebar3_hex_user.erl
+++ b/src/rebar3_hex_user.erl
@@ -86,6 +86,8 @@ format_error({generate_key, Reason}) ->
     io_lib:format("Failure generating authentication tokens: ~ts", [Reason]);
 format_error(no_match_local_password) ->
     "Password confirmation failed. The passwords must match.";
+format_error(local_password_too_big) ->
+    "Local passwords can not exceed 32 characters.";
 format_error(bad_command) ->
     "Command must be one of register, whoami, auth, deauth or reset_password";
 format_error(Reason) ->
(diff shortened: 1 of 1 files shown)


starbelly:
I will open up a PR to forward port this to master after this is merged.

ferd (approved):

PR #258, opened 2021-09-06
Improve local password collection 

starbelly:
- Resolves an issue where by if we accept a local password over 32 bytes we'll crash at key generation
  time. Instead, we now warn if the password is over 32 chars and prompt the user to try again. On the
  third attempt we throw a provider error.
 - To go along with the above, we now will warn the user if confirmation of a local password
   does not match the original entered local password vs immediately throwing, instead the user has 3 attempts
   to correctly confirm.


starbelly:
This just forward ports the commit to v6.11 branch from yesterday.

ferd (approved):

PR #269, opened 2021-12-23
Rework user and key providers

### LINKED ISSUE #203: Hex does not actually auth when running the auth task
```
> rebar3 hex user auth
Username: ([])> eproxus
Account Password: <OBVIOUSLY WRONG PASSWORD>                         
You have authenticated on Hex using your account password. However, Hex requires you to have a local password that applies only to this machine for security purposes. Please enter it.
Local Password: ^C                      
BREAK: (a)bort (A)bort with dump (c)ontinue (p)roc info (i)nfo
       (l)oaded (v)ersion (k)ill (D)b-tables (d)istribution
```

The above can't be true since I entered a bogus password on purpose.

**Environment**
```
> rebar3 --version
rebar 3.14.4 on Erlang/OTP 23 Erts 11.1.8
> rebar3 as global plugins list
--- Global plugins ---
rebar3_hex (6.10.3)
```

starbelly:
- refactor user in general to align with publish provider
- improve ux around password prompts
- move all user related key tasks to the user provider
- move encrypt_write_key and decrypt_write_key to rebar3_hex_config
- change rebar3_hex_key to become a general module instead of a provider
- adjust tests per changes

### DIFF
--- a/src/rebar3_hex.erl
+++ b/src/rebar3_hex.erl
@@ -10,7 +10,9 @@
         , help_opt/0
         ]).
 
--type task() :: #{args := map(), 
+-type task() :: #{raw_opts := proplists:proplist(),
+                  raw_args := list(),
+                  args := map(), 
                   repo := map(), 
                   apps := [rebar_app_info:t()], 
                   state := rebar_state:t(), 
@@ -22,7 +24,6 @@
 init(State) ->
     lists:foldl(fun provider_init/2, {ok, State}, [rebar3_hex_user,
                                                    rebar3_hex_cut,
-                                                   rebar3_hex_key,
                                                    rebar3_hex_owner,
                                                    rebar3_hex_repo,
                                                    rebar3_hex_search,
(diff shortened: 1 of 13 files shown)


src/rebar3_hex.erl
```diff
@@ -10,7 +10,9 @@
         , help_opt/0
         ]).
 
--type task() :: #{args := map(), 
+-type task() :: #{raw_opts := proplists:proplist(),
```
starbelly, line 13: These names don't make a lot of sense, but it's internal so it should be fine for now.

src/rebar3_hex.erl
```diff
@@ -60,16 +61,21 @@ task_args(State) ->
             {Task, proplists:delete(task, Opts)}
     end.
 
+
+%% TODO: A current limitation of rebar3_hex:task_state/1 is it returns strings.
```
starbelly, line 65: In truth, we'll probably always need to return these. `rebar3 hex retire` is a good example of where this pattern simply breaks down without the ability to define a spec for bare arguments and their expected order.

src/rebar3_hex_config.erl
```diff
@@ -27,6 +29,46 @@ api_key_name(Key, Suffix) ->
...
+
+pad(Binary) ->
+    case byte_size(Binary) of
+        Size when Size =< 16 ->
+            <<Binary/binary, 0:((16 - Size) * 8)>>;
+        Size when Size =< 24 ->
+            <<Binary/binary, 0:((24 - Size) * 8)>>;
+        Size when Size =< 32 ->
+            <<Binary/binary, 0:((32 - Size) * 8)>>
+    end.
```
ferd, line 70: huh, so would passwords >32 characters just be truncated/invalid?

ferd (approved):

src/rebar3_hex_config.erl
starbelly: No, we simply don't allow them per https://github.com/erlef/rebar3_hex/pull/269/files#diff-f91a12f48327c7c0b04d92bc558861bb13dbdbcff61f9f72f45c47ed2e8f8c1aR272

Issue #262, opened 2021-11-19
How to publish a package that depends on the package that exists in "hexpm:private_org"


voluntas:
If I'm asking the wrong question in this issue, I'd appreciate it if you could point me to the right place.

I'm trying to use private_org in hexpm using rebar3 hex.

However, when I try to publish a private_org package with it specified as a deps package in rebar.confg, it tells me that this package does not exist in hexpm and I cannot publish it.

After following the source code, I found out that this is a message returned from the hexpm side.

https://github.com/hexpm/hexpm/blob/main/lib/hexpm/repository/requirement.ex#L29

How can I publish a package that depends on a package in hexpm:private_org to hexpm:private_org?

## rebar.config

```
{deps, [
   {other_private_pkg_name, "1.0.0"}
]}.

{hex, [{repos, [#{name => <<"hexpm:private_org">>},
                #{name => <<"hexpm">>}]}]}.
```

## publish command error

```
$ rebar3 hex publish -r hexpm:private_org

Publishing private_pkg_name 1.0.0 to hexpm:private_org

===> Failed to publish package: Validation error(s)
	Requirements could not be computed
other_private_pkg_name
--------------------
package does not exist in repository "hexpm"
```

tsloughter:
This likely doesn't work and is a bug. 

But to be clear, compiling works? So it is able to fetch that private package (and I take it, publish that package as well?) and breaks only for publishing a package depending on a private package?

I'll have to look into where we do this check it is doing that results in it saying it can't find the package.

starbelly:
I'd agree it's a bug of sorts. Currently, there is a no way to declare a dependency in rebar.config as belonging to a private repo, thus when we send up the requirement without specifying a repo for the dep,  and it ends up in the global name space and why you end up with that error. 

☝️ That's an assumption without digging through hex code.

 @wojtekmach  is that assumption correct? Or should hexpm be searching for the dep in the users private's org regardless?

tsloughter:
So the error is coming from hexpm?

Question I have then is if deps can depend on deps from other private orgs?

If not, then I'd agree it sounds like something hexpm could be doing since it simply has to be in the same org you are publishing to.

If they can span multiple orgs then we need to be the ones ensuring we are publishing the package with a dependency on the exact same package (meaning specifying the org) that was being used locally.

starbelly:
> Question I have then is if deps can depend on deps from other private orgs? 

They can. 

[...]

PR #274, opened 2021-12-25
Rework repo provider

starbelly:
- align with organizations from mix hex


test/support/hex_api_model.erl
```diff
@@ -21,6 +21,33 @@
 handle(Req, _Args) ->
     handle(Req#req.method, elli_request:path(Req), Req).
 
+handle('GET', [<<"auth">>], Req) -> 
+    respond_with(200, Req, #{});
+
+handle('GET', [<<"orgs">>, <<"foo">>, <<"keys">>], Req) ->
```
starbelly, line 27: This is gross, specifically the hard code, but it'll do for now.

src/rebar3_hex_organization.erl
```diff
@@ -0,0 +1,401 @@
+%% @doc
+%% `rebar3_hex_organization' - Manage organization
+%%
+%% Manages the list of authorized hex organizations.
+%%
+%% Note that all commands that require a `NAME' argument expect a qualified repository name for the
```
starbelly, line 6: This is the only thing I'm very unsure of. If we only accept the org part of the repo name, then we have to make some assumptions. In the case you have multiple repos, which one does the org belong to? We can say this is what the `-r` switch is for, but then we have to say "In this case, you should only specify a parent repository, but in all other cases you can any repository" 

If `-r` only took parent repositories, I'd be ok with just allowing a bare name, but it doesn't, so 🤷 . I mused about switching to that. Basically, it would require that we have a `--organization` on most tasks. 

For example : 

```
rebar3 hex publish -r hexpm --organization foo
```

or

```
rebar3 hex organization -r hexpm key foo generate 
```

or 

```
rebar3 hex owner add rebar3_hex hunter42@example.com -r hexpm --organization foo
```

So, when I think about that just having `-r` and requiring the fully qualified repo name for the org task sounds quite good to me.

rebar.config
```diff
@@ -25,8 +26,7 @@
 
 {dialyzer, [
     {warnings, [
-        error_handling,
-        underspecs
```
starbelly, line 29: We'll come back and deal with this later, there's an issue around it.

src/rebar3_hex_organization.erl
ferd: I feel that ideally you should be able to reuse the same format we use when declaring dependencies, so that the notation is uniform across publishing and usage when possible?

starbelly: I agree and what this currently is. That would be `hexpm:foo` if I'm following correctly. That is, we currently require you to put in `{hex, [{repos, [ #{name => <<"hexpm:myorg">>]}]}` in rebar.config for fetching private private packages. 

In that respect, we should remove support for short hand from publishing. We currently support allowing you to say `rebar3 hex publish foo` and we will try to find the matching repository, but there's a rub in there, in the case you have an org called foo at two different parent repositories, thus a bug really, no one has run into it yet though.

starbelly: Note: We currently have no support in rebar3 (AFAIK) for specifying what repository a dep specifically should come from, this was raised in #262.

starbelly: Per this conversation, I will remove support for "shorthand" repo arguments (i.e., `-r foo`) in the upcoming config provider PR.

ferd (approved):

PR #275, opened 2021-12-26
Support --unretire on retire provider

starbelly:
- add support for --unretire
 - add doc
 - add retire/unretire helpers on rebar3_hex_client


starbelly:
Ready but depends on #274

ferd (approved):

PR #277, opened 2021-12-28
Add build provider

### LINKED ISSUE #233: Support for exclude_patterns?
Hex has the ability to exclude files from hex.pm package uploads (e.g, image files) via the `exclude_patterns` directive:
* https://hexdocs.pm/hex/Mix.Tasks.Hex.Build.html#module-package-configuration

Is this currently possible with rebar3_hex?

If not, could support for this be added?

Note that I tried adding the directive to my `src/*.app.src` file in a project but had no success ... might be doing in wrong? `git grep` and Google searches turned up nothing for rebar3 + hex.

starbelly:
This is ready but depends on #272 being merged first.

Closes #233

### DIFF
--- a/src/rebar3_hex_build.erl
+++ b/src/rebar3_hex_build.erl
@@ -329,10 +329,11 @@ include_files(Name, AppDir, AppDetails) ->
     FilePaths = proplists:get_value(files, AppDetails, ?DEFAULT_FILES),
     %% In versions prior to v7 the name of the for including paths and excluding paths was include_files and
     %% exclude_files. We don't document this anymore, but we do support it to avoid breaking changes. However, 
-    %% users should be instructed to use *_paths.
+    %% users should be instructed to use *_paths. Likewise for exclude_regexps which is now documented as
+    %% exclude_patterns.
     IncludePaths = proplists:get_value(include_paths, AppDetails, proplists:get_value(include_files, AppDetails, [])),
     ExcludePaths = proplists:get_value(exclude_paths, AppDetails, proplists:get_value(exclude_files, AppDetails, [])),
-    ExcludeRes = proplists:get_value(exclude_patterns, AppDetails, []),
+    ExcludeRes = proplists:get_value(exclude_patterns, AppDetails, proplists:get_value(exclude_regexps, AppDetails, [])),
 
     AllFiles = lists:ukeysort(2, rebar3_hex_file:expand_paths(FilePaths, AppDir)),
     IncludeFiles = lists:ukeysort(2, rebar3_hex_file:expand_paths(IncludePaths, AppDir)),
--- a/src/rebar3_hex_build.erl
+++ b/src/rebar3_hex_build.erl
@@ -45,10 +45,10 @@
 %%  <li>`links' - A map where the key is a link name and the value is the link URL. Optional but highly recommended.
 %%  <li> `files' - A list of files and directories to include in the package. Defaults to standard project directories, 
 %%        so you usually don't need to set this property.</li>
-%%  <li> `include_files' - A list of paths containing files you wish to include in a release. </li>
-%%  <li> `exclude_files' - A list of paths containing files you wish to exclude in a release. </li>
+%%  <li> `include_paths' - A list of paths containing files you wish to include in a release. </li>
+%%  <li> `exclude_paths' - A list of paths containing files you wish to exclude in a release. </li>
 %%  <li> `exclude_patterns' - A list of regular expressions used to exclude files that may have been accumulated via
-%%  `files' and `include_files' and standard project paths. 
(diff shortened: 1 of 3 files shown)


src/rebar3_hex_build.erl
```diff
@@ -1,3 +1,88 @@
...
+%% In addition, the following meta attributes are supported and highly recommended : 
+%%
+%% <ul>
+%%  <li> `description' - a brief description about your application.
+%%  <li>`pkg_name' - The name of the package in case you want to publish the package with a different name than the
+%%  application name.</li>
+%%  <li>`links' - A map where the key is a link name and the value is the link URL. Optional but highly recommended.
+%%  <li> `files' - A list of files and directories to include in the package. Defaults to standard project directories, 
+%%        so you usually don't need to set this property.</li>
+%%  <li> `include_files' - A list of paths containing files you wish to include in a release. </li>
```
starbelly: I feel like these should be called `include_paths` and `exclude_paths` and we could also perhaps support `exclude_files`. They both work on paths right now. Thoughts?

src/rebar3_hex_build.erl
```diff
@@ -1,3 +1,88 @@
...
+%%                   stdlib,
+%%                   ]},
+%%   {licenses, ["Apache-2.0"]},
+%%   {links, [{"GitHub", "https://github.com/my_name/myapp"}]}]}.
+%% ```
+%%
+%% <h2> Command line options </h2>
+%%
+%% <ul>
+%%  <li>`--repo' - Specify the repository to use in the task. This option is required when
```
starbelly: I don't like that we support this on this provider, but we currently support being able to configure a doc provider on a per repo basis.

src/rebar3_hex_build.erl
ferd: the `_paths` variants are clearer, but it's not a super significant difference either way, especially if this ends up using glob or wildcard patterns (though that's `exclude_patterns`, which has also been renamed). The "confusion" is already common (e.g. `filelib` supports `is_file|dir|regular` already). 

I'd tend to still lower the updating friction for the plugin by keeping this compatible with the stuff that was already in the .app.src file though, to make upgrading the plugin as easy as possible. The 7.0 update will likely have people update their licenses, so there should at least be an obvious way to know that you're done fixing what broke rather than suddenly having older options not work (if any, I mostly recall just using `files`).

Either way I'm fine with the direction we're going.

src/rebar3_hex_build.erl
```diff
@@ -1,3 +1,88 @@
...
+%%   {licenses, ["Apache-2.0"]},
+%%   {links, [{"GitHub", "https://github.com/my_name/myapp"}]}]}.
+%% ```
+%%
+%% <h2> Command line options </h2>
+%%
+%% <ul>
+%%  <li>`--repo' - Specify the repository to use in the task. This option is required when 
+%%      you have multiple repositories configured, including organizations. The argument must 
+%%      be a fully qualified repository name (e.g, `hexpm', `hexpm:my_org', `my_own_hexpm').
```
ferd: PRobably worth adding <code>defaults to `expm`</code>

src/rebar3_hex_build.erl
```diff
@@ -1,3 +1,88 @@
...
+%%   </li>
+%%   <li> `--unpack' - Builds the tarball and unpacks contents into a directory. Useful for making sure the tarball 
+%%        contains all needed files before publishing. See --output below for setting the output path.
+%%   </li>
+%%   <li> `-o', `--output' - Sets output path. When used with --unpack it means the directory 
+%%   (Default: <app>-<version>). Otherwise, it specifies tarball path (Default: <app>-<version>.tar)</li>
```
ferd: We would usually tend to default the output to somewhere in `_build/<profile>/` so these don't get to bypass .gitignore on build artifacts.

starbelly: Good point. Will change.

starbelly:
> the _paths variants are clearer, but it's not a super significant difference either way, especially if this ends up using glob > or wildcard patterns (though that's exclude_patterns, which has also been renamed). The "confusion" is already common (e.g. filelib supports is_file|dir|regular already).

I'll add support for `_paths` but keep in support for the old name.

ferd:
yeah just supporting newer options without breaking the old stuff is fine, if it's not too much extra complexity. Keep a comment about the reason for it so whoever is still maintaining this in the future has a good contextual clue of the intent behind it.

src/rebar3_hex_build.erl
starbelly: Done.

ferd (approved):

starbelly:
Note : I made one more change, in line with the others around name changes (i.e., continue to support exclude_regexps).

PR #285, opened 2021-12-30
Only expand once

### LINKED ISSUE #246: Infinite recursion bug in rebar3_hex_file:expand_paths/2
A bug was raised on erlanger slack by @eproxus. It's possible with the perfect storm of symlinks to end up an infinite loop. This happens at https://github.com/erlef/rebar3_hex/blob/master/src/rebar3_hex_file.erl#L21 . Technically you could say there's a bug in OTP, but that's another story. Instead we need to change the way we discover files. I think we should probably do something similar to what mix hex is doing here https://github.com/hexpm/hex/blob/cf0b48d09b31312c574ec3ad57ef69278e66ff46/lib/mix/tasks/hex.build.ex#L359 

The change should include tests to ensure we only expand paths that are given to `expand_paths/2`. Currently per the double star wildcard call, we end up expanding outside of those paths. 

Diagnostics should be added as well such that if the new way of expanding paths and finding files the operator would get some indication that this is happening and where.

starbelly:


starbelly:
I need to add a regression test for this at the very least. Would really like @eproxus to give this a whirl himself as well. 

I tested this against the edge case kraft tarball I was given and no more infinite recursion, but would still like further validation (beyond unit tests). 


To test this simple go to a project, recreate the scenario per #246 and run `mkdir -p tmp && rebar3 hex build package -o tmp` 

Then inspect the goods.

src/rebar3_hex_file.erl
```diff
@@ -1,34 +1,100 @@
...
-    rebar_dir:make_relative_path(AbsFile, AbsDir).
-
 update_app_src(App, Version) ->
     AppSrcFile = rebar_app_info:app_file_src(App),
     AppSrc = rebar_file_utils:try_consult(AppSrcFile),
     [{application, Name, Details}] = AppSrc,
     NewDetails = lists:keyreplace(vsn, 1, Details, {vsn, rebar_utils:to_list(Version)}),
     {application, Name, NewDetails}.
+
+relative_to(Path, From) ->
```
starbelly: I tried keeping `rebar_dir:make_relative_path/1` but ran into problems. Can look into that again, but I figured eh, just take the implementation from Elixir's Path since the algo is based on what's going on in hex anyway.

starbelly:
Right, coming back to this after a good while, it's not infinite recursion, but if you have a lot of files, it can take a LONG time in addition to ending up with a ton of bogus paths in the results.

src/rebar3_hex.erl
```diff
@@ -11,6 +11,8 @@
         , help_opt/0
         ]).
 
+-export([to_atom/1, to_list/1, to_binary/1]).
```
starbelly, line 14: Rationale : the less we rely on rebar3 internal util modules, the less burden it is for rebar3 to make changes without worrying about breaking the rest of the world.

src/rebar3_hex_file.erl
```diff
@@ -1,35 +1,93 @@
...
+-export([
+    expand_paths/2,
+    update_app_src/2
+]).
 
 expand_paths(Paths, Dir) ->
-    AbsDir = filename:absname(Dir),
-    Files = lists:flatmap(fun dir_files1/1, [filename:join(Dir, P) || P <- Paths]),
-    [{relative_path(F1, AbsDir), F1} || F1 <- filter_regular(Files)].
+    AbsDir = expand(rebar3_hex:to_binary(Dir)),
```
starbelly, line 12: As mentioned this is pretty much a 1:1 with the mix hex algo and the path related functions are pretty much a 1:1 with utilities from stdlib in elixir. I think this path makes sense, especially considering I'd like to reach the point to where most of the time we working with binaries vs strings, when possible. 

In mix hex, there is no filtering of regular files either, directory paths are left in the list as is, as an example, which makes sense to me.  In addition, symlinks are allowed in packages.

src/rebar3_hex_build.erl
```diff
@@ -456,18 +456,13 @@ create_docs(State, Repo, App, Args) ->
...
                 true ->
-                    AppDir = rebar_app_info:dir(App),
                     AppDetails = rebar_app_info:app_details(App),
-                    Files = rebar3_hex_file:expand_paths([DocDir], AppDir),
+                    Files = rebar3_hex_file:expand_paths([DocDir], DocDir),
                     Name = rebar_utils:to_list(rebar_app_info:name(App)),
                     PkgName = rebar_utils:to_list(proplists:get_value(pkg_name, AppDetails, Name)),
                     OriginalVsn = rebar_app_info:original_vsn(App),
                     Vsn = rebar_utils:vcs_vsn(App, OriginalVsn, State),
-                    FileList = [
```
starbelly, line 466: This became redundant (and broken) after modifying rebar3_hex_file if we simply change the dir paths should be relative to the docs dir.

test/rebar3_hex_build_SUITE.erl
```diff
@@ -63,8 +67,18 @@ create_package_with_alias_deps(Config) ->
 create_docs_test(Config) ->
     StubConfig = #{type => app, dir => data_dir(Config), name => "valid"},
     {State, Repo, App} = test_utils:make_stub(StubConfig),
-    {ok, _Docs} = rebar3_hex_build:create_docs(State, Repo, App),
-    ok.
+    {ok, #{tarball := Tarball}} = rebar3_hex_build:create_docs(State, Repo, App),
```
starbelly, line 70: Added a test to ensure we don't create docs tarball with files insides a "doc" dir.

test/rebar3_hex_file_SUITE.erl
```diff
@@ -15,9 +15,23 @@ expand_paths_test(_Config) ->
     File = string:join([Dir2, "bar"], "/"),
     ok = file:write_file(File, ""),
     ?assertEqual([{"bar", File}], rebar3_hex_file:expand_paths(["bar"], "foo")),
-    Exp = [{"foo/bar", File}],
+    Exp = [{"foo", Dir2}, {"foo/bar", File}],
     ?assertEqual(Exp, rebar3_hex_file:expand_paths(["foo*"], Dir)).
 
+expand_once_test(_Config) ->
```
starbelly, line 21: There's one ugly truth that effects both mix hex and rebar3 hex, that is if someone gives a double star glob (i.e., `ouch/**`). We could guard against it, perhaps but we would invalidate other "safe" cases.

None the less with the previous implementation we would end up with : 

```
["/Users/starbelly/devel/erlang/truecoat/_checkouts/rebar3_hex/ouch/regression",
 "/Users/starbelly/devel/erlang/truecoat/_checkouts/rebar3_hex/ouch/regression/foo",
 "/Users/starbelly/devel/erlang/truecoat/_checkouts/rebar3_hex/ouch/regression/foo/bar",
 "/Users/starbelly/devel/erlang/truecoat/_checkouts/rebar3_hex/ouch/regression/foo/bar/baz",
 "/Users/starbelly/devel/erlang/truecoat/_checkouts/rebar3_hex/ouch/regression/foo/bar/baz/regression",
 "/Users/starbelly/devel/erlang/truecoat/_checkouts/rebar3_hex/ouch/regression/foo/bar/baz/regression/foo",
 "/Users/starbelly/devel/erlang/truecoat/_checkouts/rebar3_hex/ouch/regression/foo/bar/baz/regression/foo/bar",
 "/Users/starbelly/devel/erlang/truecoat/_checkouts/rebar3_hex/ouch/regression/foo/bar/baz/regression/foo/bar/baz",
 "/Users/starbelly/devel/erlang/truecoat/_checkouts/rebar3_hex/ouch/regression/foo/bar/baz/regression/foo/bar/baz/regression",
 "/Users/starbelly/devel/erlang/truecoat/_checkouts/rebar3_hex/ouch/regression/foo/bar/baz/regression/foo/bar/baz/regression/foo",
...
 ```

ferd (approved):

PR #286, opened 2021-12-30
Fixup unhandled errors in new build provider

starbelly:
Caught when testing new expand_paths/2 in kraft

### DIFF
--- a/src/rebar3_hex_build.erl
+++ b/src/rebar3_hex_build.erl
@@ -107,7 +107,7 @@
     "NOTICE"
 ]).
 
--define(DEPS, [{default, lock}]).
+-define(DEPS, [{default, compile}, {default, lock}]).
 -define(PROVIDER, build).
 -define(DEFAULT_DOC_DIR, "doc").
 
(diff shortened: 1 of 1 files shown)


src/rebar3_hex_build.erl
```diff
@@ -236,12 +264,22 @@ write_or_unpack(App, #{type := Type, tarball := Tarball, name := Name, version :
     end,
     AbsOut.
         
+%% We are exploiting a feature of ensuredir that that creates all
+%% directories up to the last element in the filename, then ignores
+%% that last element. This way we ensure that the dir is created
+%% and not have any worries about path names
 output_dir(App, #{output_dir := undefined}) ->
-    rebar_app_info:out_dir(App);
+    Dir = filename:join([rebar_app_info:out_dir(App), "hex"]),
```
starbelly, line 272: Also, did this here because if someone goes into the dir returned by `out_dir/1` and tries to unpack everything they'll end up with an error per trying to unpack into symlinks.

ferd (approved):

PR #297, opened 2021-12-31
Expect ok to be returned if there is no docs config in hex config

starbelly:


rebar.config
```diff
@@ -40,6 +40,7 @@
 
 {alias, [{test, [{ct, "--cover"}, {cover, "-v"}]}]}.
 
+{hex, [{doc, ex_doc}]}.
```
starbelly, line 43: Also forgot this to be set.

ferd (approved):

PR #303, opened 2022-03-28
Don't require repo arg for building under certain conditions

### LINKED ISSUE #299: build command should not require repo argument
We are using `rebar3 hex build` command to build packages for our self-hosted hex repo, but since we have multiple repos defined in rebar.config, `rebar3 hex build` complains that "A repository argument is required", but since build command only builds the package and doesn't publish it to the repo, I think this repo argument can be removed from build command, right?

starbelly:
Most cases for using the build provider shouldn't require specifying a repo argument.
The exception is when there is more than one repo found and at least one has doc config set on the
repo.

### DIFF
--- a/src/rebar3_hex_build.erl
+++ b/src/rebar3_hex_build.erl
@@ -1,34 +1,34 @@
-% @doc `rebar3 hex build' - Build packages and docs 
+% @doc `rebar3 hex build' - Build packages and docs
 %
 %% Builds a new local version of your package.
 %%
-%% By default this provider will build both a package tarball and docs tarball. 
-%%  
-%% The package and docs .tar files are created in the current directory, but is not pushed to the repository. An app 
-%% named foo at version 1.2.3 will be built as foo-1.2.3.tar. Likewise the docs .tar would be built as 
-%% foo-1.2.4-docs.tar. 
+%% By default this provider will build both a package tarball and docs tarball.
+%%
+%% The package and docs .tar files are created in the current directory, but is not pushed to the repository. An app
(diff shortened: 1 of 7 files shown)


src/rebar3_hex_build.erl
```diff
@@ -171,49 +171,84 @@ init(State) ->
...
+             case Repos of
+                 [Repo] ->
+                     Repo;
+                 [_|_] ->
+                     {Opts, _} = rebar_state:command_parsed_args(State),
+                     case proplists:get_value(task, Opts, undefined) of
+                         Task when Task =:= undefined orelse Task =:= "docs" ->
+                             case has_no_doc_provider(State, Repos) of
+                                 true ->
+                                     #{name => repo_not_set};
```
ferd: What's up with this magic value?

starbelly: name could be anything, but we have a clause for doc_opts/1 that matches on a map with name in it. I figured if we crash with passing this around, there's only one place it would be set.

starbelly: Note : I'm going to adjust this, it's hacky.

starbelly:
@ferd  I squashed and forced pushed

1. Got rid of the hack in favor of just using the default hexpm repo which aligns with our current documentation.
2. Don't do a bunch of checks to see if there are any repos with doc config, instead if you want to use repo with doc config, you must specify it. 
3. Added deprecation warning if a repo is used with doc config

ferd (approved):

PR #308, opened 2022-04-10
Warn on prompt with instructions if no doc provider is configured

starbelly:
Screen Shot 2022-04-10 at 5 13 26 PM

This is quite verbose, maybe too verbose, but I figured giving some helpful instructions and being a little noisy is a good way to go here. Especially since this will become an error in 7.1. 

Edit: 

The only other thing I think we can add is link to docs for rebar3_hex.

### DIFF
--- a/src/rebar3_hex_publish.erl
+++ b/src/rebar3_hex_publish.erl
@@ -385,6 +385,7 @@ handle_task(_) ->
 
 -dialyzer({nowarn_function, publish/4}).
 publish(State, Repo, App, Args) ->
+    maybe_warn_about_doc_config(State, Repo),
     HexConfig = rebar3_hex_config:get_hex_config(?MODULE, Repo, write),
     case publish_package(State, HexConfig, App, Args) of
         abort ->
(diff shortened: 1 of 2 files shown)


src/rebar3_hex_publish.erl
```diff
@@ -492,6 +493,38 @@ maybe_prompt(_Args, Message) ->
...
+                      "you should use : ~n~nrebar3 hex publish package~n~n"
+                      "Note : In version 7.1 this warning will be treated as an error~n",
+                      rebar_api:warn(Warning, [
+                                               "{ex_doc, [", 
+                                               " ", 
+                                               "{source_url, <<\"https://github.com/namespace/your_app\">>},",
+                                               " ",
+                                               "{extras, [<<\"README.md\">>, <<\"LICENSE.md\">>]},",
+                                               " ",
+                                               "{main, <<\"readme\">>}]}"
```
ferd:
```suggestion
                                               "{main, <<\"readme\">>}]}."
```

ferd (approved):

max-au:
I am probably running into some different error.
When I run `rebar3 hex publish docs` in the root of the https://github.com/max-au/erlperf repository, I get this error:
```
===> docs argument given, will not publish package
===> No documentation provider was configured, docs will not be generated.
```

(that is rebar3 from the master branch, fresh from the oven).

I don't exactly understand why adding `--doc-root=doc` fixes the problem. But it does:
```
rebar3 hex publish docs --doc-dir=doc
===> docs argument given, will not publish package
Local Password:                         
===> Published docs for erlperf 2.0.1   
```

starbelly:
@max-au It's the same issue. You are required to configure your doc provider. If you're using `edoc` then it should be `edoc`, if `rebar3_ex_doc`, ... 

The reason `--doc-dir` would work is if you generated your docs using `edoc` outside of rebar3_hex. `--doc-dir` to support cases where there is no doc provider (i.e. not using edoc or rebar3_ex_doc). 

#308 attempts to warn the user ahead of time, in a later version this will become an error.

Issue #316, opened 2023-01-25
Crash while running `rebar3 hex cut`


saleyn:
## Description ##

When running `rebar3 hex cut` in the source of the [erlcron](https://github.com/erlware/erlcron) repository, I get a rebar3.crashdump.  

```
Error: function_clause
[{rebar3_hex_cut,cut,
  [{state_t,"/home/serge/projects/erl-libs/erlcron",
    {dict,28,16,16,8,80,48,
     {[],[],[],[],[],[],[],[],[],[],[],[],[],[],[],[]},
     {{[],
       [[{deps,default}],
  ...
    0,undefined,"/home/serge/projects/erl-libs/erlcron",
    "/home/serge/projects/erl-libs/erlcron/_build/default/lib/erlcron",
    "/home/serge/projects/erl-libs/erlcron/_build/default/lib/erlcron/ebin",
    undefined,false,false,false,undefined,false},
   #{task => undefined}],
  [{file,
    "/home/serge/projects/erl-libs/erlcron/_build/default/plugins/rebar3_hex/src/rebar3_hex_cut.erl"},
   {line,141}]},
 {lists,foreach_1,2,[{file,"lists.erl"},{line,1442}]},
 {rebar3_hex_cut,handle_task,1,
  [{file,
    "/home/serge/projects/erl-libs/erlcron/_build/default/plugins/rebar3_hex/src/rebar3_hex_cut.erl"},
   {line,122}]},
 {rebar_core,do,2,
  [{file,"/home/serge/projects/erl-libs/rebar3/apps/rebar/src/rebar_core.erl"},
   {line,155}]},
 {rebar_prv_do,do_task,5,
  [{file,
    "/home/serge/projects/erl-libs/rebar3/apps/rebar/src/rebar_prv_do.erl"},
   {line,87}]},
 {rebar_core,do,2,
  [{file,"/home/serge/projects/erl-libs/rebar3/apps/rebar/src/rebar_core.erl"},
   {line,155}]},
 {rebar3,run_aux,2,
  [{file,"/home/serge/projects/erl-libs/rebar3/apps/rebar/src/rebar3.erl"},
   {line,193}]},
 {rebar3,main,1,
  [{file,"/home/serge/projects/erl-libs/rebar3/apps/rebar/src/rebar3.erl"},
   {line,66}]}]
```

Here's the `rebar.config`:
```
$ cat rebar.config 
%% -*- mode: Erlang; fill-column: 80; comment-column: 75; -*-
[...]

PR #317, opened 2023-01-25
Fix  issue with `hex cut` when no `-i` option is given (#316)

saleyn:
The `-i` (increment) option is documented as optional but when not passed, it crashes `hex cut`. This fix makes it optional.


ferd (approved):
Looks valid to me. Broken CI unrelated it seems.

starbelly:
This looks good, CI failure is unrelated but cut currently isn't covered anyway, so will YOLO merge this one. Thanks! ❤️

PR #326, opened 2023-03-11
Add option to opt-out of v prefix on tags / commits

starbelly:
This allows users to opt of of the `v` prefix for git tags and commits.

### DIFF
--- a/src/rebar3_hex_cut.erl
+++ b/src/rebar3_hex_cut.erl
@@ -10,6 +10,8 @@
 %% `cut' will also optionally create version bump commit, create a tag with a name corresponding to the new version, and
 %%  prompt you if you'd like to push the tag up to your git repository.
 %%
+%% You may opt-out of prefixing tags and commits with a `v' prefix (i.e., v1.2.3) using
+%% `--prefix-tag false' switch.
 %%
 %% Below is a full example of `cut' in action :
 %%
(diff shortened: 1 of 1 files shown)


src/rebar3_hex_cut.erl
```diff
@@ -98,6 +100,8 @@ init(State) ->
                                 {desc, ""},
                                 {opts, [{increment, $i, "increment", {string, undefined},
                                          "Type of semver increment: major, minor or patch"},
+                                        {prefix_tag, $p, "prefix-tag", {boolean, true},
```
starbelly, line 103: Maybe this should be called `prefix-source-ref`  or `prefix-ref`

src/rebar3_hex_cut.erl
```diff
@@ -141,39 +145,46 @@ format_error(Reason) ->
...
                 {error, publish_failed} ->
-                    delete_tag(NewVersion),
+                    delete_tag(MaybePrefixedVersion),
                     {ok, State}
             end
 
     end.
 
+maybe_prefix_version(Vsn, true) ->
+    io_lib:format("v~s?", [Vsn]);
```
tsloughter: Is that `?` a typo?

starbelly: It would be. Fixed.

tsloughter (approved):

PR #328, opened 2023-07-01
Fix authentication to private repos `rebar3 hex user auth --repo [REPO]`

derekkraan:


starbelly:
Hey @derekkraan can you provide steps to reproduce the issue?

derekkraan:
Hey, @starbelly, yes. This is the short version since I'm just on mobile now. I am working on a private repo for hosting (among others) Erlang packages. Running `rebar3 hex user auth --repo codecodeship` results in the auth data being saved for hexpm (in the rebar3 state) and not for code code ship, resulting in further calls (such as `rebar3 hex publish --repo codecodeship`) failing due to missing keys.

I hope this makes sense. I can do a better write-up on Monday when I'm back in front of a computer if necessary. Also happy to answer any follow up questions.

starbelly:
@derekkraan Ahhh I see, I literally just reproduced the issue right before you replied. This LGTM!

starbelly:
This bug fix reveals some other latent bugs. 

1. We're not setting api_organization to undefined in fetching / building hex config at runtime (line 241 of rebar3_hex_config.erl)
2. We're not setting api_repository for non-hexpm config, this is included by default for hexpm default config, but not so for this case. This is probably per testing with nothing but hexpm and in that referring to it as hexpm vs an alternative name. 

Of course, there's a fun dialyzer error to fix as well :) I'm happy you have opened this PR though so we can solidify this use case!

derekkraan:
@starbelly there seems to be some confusion in the codebase. I believe that dialyzer is complaining about the attribute `name`, which doesn't exist on the type `config()` found in `apps/rebar/src/vendored/r3_hex_core.erl` (only `repo_name`). I am not really sure what to do about this, since rebar3_hex is using `name` all over the place. Is this a bug? Do I need to set both `name` and `repo_name` in `rebar.config`?

I added some code to make sure that `api_organization` and `api_repository` are both being populated with `undefined` when that is appropriate.

derekkraan:
Setting `name` in `rebar.config` appears to be enough to also set `repo_name`. I think this PR is good to go.

`{hex, [{repos, [#{name => <<"codecodeship">>, api_url => <<"https://hex.codecodeship.com/api">>}]}]}.`

starbelly:
@derekkraan LGTM! (for real), this helped uncover yet another bug, but I'll fix that up in another PR. Thank you!

starbelly (approved):

PR #334, opened 2024-03-17
A better TLDR of what's rebar3_hex [skip ci]

### LINKED ISSUE #333: Is rebar3_hex required in modern/recent versions of rebar3?
While reading the [docs on pkg management at rebar3.org](https://rebar3.org/docs/package_management/), the subcmd `hex` is presented as a built-in thing instead of a separate plugin that needs to be installed before usage. I'm not very familiar with rebar3, so I'm not sure if I need to "install" rebar3_hex in my `rebar3.config` or if it's something that comes bundled in modern rebar3.

I'm using Erlang v25.3.2.3 and rebar3 v3.22.1, both installed through [asdf](https://asdf-vm.com/).

Whichever the answer, I believe it would be great to have it stated in `README.md`, perhaps also describing which versions of rebar3 would require installing it manually, and which ones would bring rebar3_hex bundled in itself.

joeljuca:
Given that Rebar3 provides built-in support for a limited set of features related to Hex, and there has been multiple cases of confused developers looking to use rebar3_hex for basic usage (aka.: no package publishing), I  reworded the first paragraph (TLDR) of what is the project – along with a note of when it's required.

> _"The rebar3_hex plugin is required for publishing packages only. (...)"_  
> @tsloughter. https://github.com/erlef/rebar3_hex/issues/333#issuecomment-1971076725

Closes #333

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,10 @@
 [![Docs](https://img.shields.io/badge/hex-docs-green.svg?style=flat)](https://hexdocs.pm/rebar3_hex)
 [![Erlang Versions](https://img.shields.io/badge/Supported%20Erlang%2FOTP-22.0%20to%2024.0-blue)](http://www.erlang.org)
 
-rebar3_hex is a rebar3 plugin which bundles providers for interacting with the Erlang ecosystem package manager [hex.pm](https://hex.pm/).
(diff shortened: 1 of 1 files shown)


starbelly (approved):
LGTM

PR #335, opened 2024-06-11
Bump hex_core to v0.10.1

starbelly:
Updates to the latest hex_core. Oddly enough, a spec in hex_core that has never caused problems before caused problems today. The spec is here : https://github.com/hexpm/hex_core/blob/8a53ac8eddaf425cd70a41ae9279318063c01e67/src/hex_api_release.erl#L19 

Not sure why now, but I'll open an issue in relevant places to sort it out.


paulo-ferraz-oliveira:
> a spec in hex_core that has never caused problems

The last CI I see for this is 8 months old. 24.3.4.17 is 2 months old. Might it be Erlang/OTP that changed something that caused that issue?

**Edit**: maybe it'd also be a nice time to include 26 and 27 in CI? (if you prefer, though, I can do a separate pr for that); and potentially, while we're at it, `actions/checkout` to v4, as well as `codecov/codecov-action` to v4, and also `rebar3` to 3.23.0; yeah, I'll do a separate pull request for this. --> https://github.com/erlef/rebar3_hex/pull/336

rebar.config
```diff
@@ -11,13 +11,13 @@
...
 
-{deps, [{hex_core, "0.8.4"}, {verl, "1.1.1"}]}.
+{deps, [{hex_core, "0.10.1"}, {verl, "1.1.1"}]}.
 
 {profiles, [
     {test, [
         {extra_src_dirs, ["test/support"]},
         {overrides, [{override, rebar3,[{deps, [{erlware_commons, "1.3.1"}]}]}]},
-        {deps, [{hex_core, "0.8.4"},
+        {deps, [{hex_core, "0.10.1"},
```
paulo-ferraz-oliveira, line 20: I'd get rid of this one. Doesn't `rebar3` just accumulate deps in profiles (i.e. concatenate them with whatever's in the "default" profile)?

paulo-ferraz-oliveira:
@starbelly, I'm also gonna remove `-dialyzer` directives from this repo. and fix the issues as I go along (**edit**: https://github.com/erlef/rebar3_hex/pull/338). Some of them might come from `hex_core` (at least one spec. is wrong) and I'll post a version in a pull request that uses another `hex_core` I'll pull request to fix 😄 --> https://github.com/hexpm/hex_core/pull/148

rebar.config
starbelly: I suspect this was put in play to test a different version from what rebar3 had, but in rebar3 it is now vendored.

ferd: Agreed, this should be possible to take out.

ferd (approved):

paulo-ferraz-oliveira (approved):
Maybe you'll want to remove `hex_core` from the `test` profile, but let that not affect merging+moving on. 👍

paulo-ferraz-oliveira:
Once this one's merged I'll go ahead and rebase the others.

starbelly:
> Maybe you'll want to remove `hex_core` from the `test` profile, but let that not affect merging+moving on. 👍

Yes, let's get this merged and open up a tidy PR as a follow on, we also should modify CI to test on OTP 26/27 and stop testing 24.

paulo-ferraz-oliveira:
Good stuff. I already have a couple PRs lined up. I didn't remove '25' from tests, since "current-2" seems to be acceptable for OTP versions under test. At the same time, though, in one of the PRs, I did remove `hex_core` from the `test` profile, as cleanup. 😄

PR #105, opened 2019-02-19
Fixups for publishing, config, and repo commands

starbelly:
- change instances of repos_key to repo_key (typo)
  - write repo_key to config vs auth_key. auth_key is an implementation detail of the mix hex client. 
  - Use the org name vs full repo name when generating a repo key
  - Honor the -r flag in the publish command

PR #338, opened 2024-06-12
Fixes for Dialyzer

paulo-ferraz-oliveira:
I start by removing all `-dialyzer` directives and then issuing fixes for the found issues.

I'll add self-review comments where appropriate, to ease review.


paulo-ferraz-oliveira:
> `===> Error in dialyzing apps: Unknown dialyzer warning option: no_unknown`

This is a bummer: I'ma fix CI so it only runs Dialyzer on a specific OTP version.

.github/workflows/ci.yml
```diff
@@ -13,8 +13,20 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        otp: ['24', '25']
-        rebar3: ['3.16.1']
+        include:
+          - otp: 27
+            os: ubuntu-24.04
+            rebar3: 3.23
+            dialyzer: true
```
paulo-ferraz-oliveira, line 20: Lemme know if this is acceptable. The reason for this exception is that with e.g. OTP 25 Dialyzer fails by not knowing option `no_unknown`.

rebar.config
```diff
@@ -9,9 +9,9 @@
     {platform_define, "^[2-9]", 'POST_OTP_18'}
 ]}.
 
-{project_plugins, [covertool, rebar3_ex_doc, {rebar3_hank, "~> 0.3.0"}]}.
+{project_plugins, [covertool, rebar3_ex_doc, {rebar3_hank, "1.4.0"}]}.
```
paulo-ferraz-oliveira: This is already proposed in another pull request that'll hopefully get merged before this one. Without this bump `katana-code` compilation fails on directive processing.

rebar.config
```diff
@@ -9,9 +9,9 @@
...
 
-{deps, [{hex_core, "0.8.4"}, {verl, "1.1.1"}]}.
+{deps, [{hex_core, {git, "https://github.com/paulo-ferraz-oliveira/hex_core", {branch, "fix/spec"}}}, {verl, "1.1.1"}]}.
```
paulo-ferraz-oliveira: This is obviously just to show that with this we have ✅  CI. I don't expect to keep this change and/or maintain that branch 😄

src/rebar3_hex_client.erl
```diff
@@ -96,8 +96,8 @@ delete_docs(Config, Name, Version) ->
     response(Res).
 
 retire(Config, Package, Version, Reason, Message) -> 
-    Msg = #{<<"reason">> => Reason,
-            <<"message">> => Message
+    Msg = #{reason => Reason,
+            message => Message
```
paulo-ferraz-oliveira: Maybe this wasn't throwing an exception / Dialyzer issue because somehow the keys are coerced into atoms/binaries remotely, but this is the way the spec is telling it should be.

src/rebar3_hex_organization.erl
```diff
@@ -213,20 +213,20 @@ format_error({get_repo_by_name, {error,{not_valid_repo,ParentName}}}) ->
...
         case proplists:get_value(key, Opts, undefined) of
             undefined ->
-                Config = rebar3_hex_config:get_hex_config(?MODULE, ParentRepo, write),
+                Config = rebar3_hex_config:get_hex_config(?MODULE, ParentRepo0, write),
                 Config1 = Config#{api_organization => OrgName},
                 KeyName = proplists:get_value(key_name, Opts, rebar3_hex_config:repos_key_name()),
                 generate_key(Config1, KeyName, default_perms(OrgName));
             ProvidedKey ->
                 TestPerms = #{domain => <<"repository">>, resource => OrgName},
+                ParentRepo = maps:without([name], ParentRepo0),
```
paulo-ferraz-oliveira: The spec doesn't mention name, and I'm also not sure why it's there, but it is used internally, by `rebar3_hex` though. Making this call fixes the Dialyzer issue and gets us closer to the current spec.

src/rebar3_hex_publish.erl
```diff
@@ -416,6 +414,7 @@ publish_package(State, Repo, App, Args) ->
                     rebar_api:info("--dry-run enabled : will not publish package.", []),
                     {ok, State};
                 _ ->
+                    Repo = maps:without([name], Repo0),
```
paulo-ferraz-oliveira: The reason for this change was already pointed out above.

src/rebar3_hex_retire.erl
```diff
@@ -126,8 +126,8 @@ format_error(Reason) ->
 
 retire(State, PkgName, Version, Repo, RetireReason, RetireMessage) ->
     HexConfig = rebar3_hex_config:get_hex_config(?MODULE, Repo, write),
-    Msg = #{<<"reason">> => RetireReason,
-                <<"message">> => RetireMessage},
+    Msg = #{reason => RetireReason,
+            message => RetireMessage},
```
paulo-ferraz-oliveira: The reason for this change was already pointed out above.

src/rebar3_hex_organization.erl
paulo-ferraz-oliveira: I don't know if it's preferable to keep as-was (but not respecting the spec.) or to use some other key where `name` is used...

rebar.config
paulo-ferraz-oliveira: `hex_core` 0.10.2 is out, so I've imported it here, instead of this workaround.

.github/workflows/ci.yml
starbelly: I think it's absolutely fine. There's no need to check specs on everything ;)

src/rebar3_hex_client.erl
starbelly: It doesn't work for some reason, I did an integration test locally using hexpm from main. 

Calling on @wojtekmach for clarity, maybe I was holding it wrong? I can try again.

paulo-ferraz-oliveira: Ah, then the problem might be that the specs are incorrect, in which case the code was Ok. (and not under some coercion as I'd initially thought)

paulo-ferraz-oliveira: @starbelly, might this also indicate we're missing an important test somewhere?

starbelly: @paulo-ferraz-oliveira We'd have to have a full blown integration test, which is doable.

paulo-ferraz-oliveira: I'm Ok with identifying it as an issue and postponing it, though 😉

paulo-ferraz-oliveira: Yeah, so I peaked further and it appears (as you noticed) that `hexpm` requires that as a string. The spec says `atom` which is incorrect.

I see several solutions, among which:

- drop the fine-grained spec (because you can't spec-ify a map's binary keys)
- accept atoms and convert them to binaries
- add a `-dialyzer` directive (dirtier) and move on

src/rebar3_hex_organization.erl
paulo-ferraz-oliveira: @starbelly, do you know where `name` comes from? Is it potentially just a quick local hack? Are you Ok with the current solution of removing it to obey the spec?

starbelly: `name` comes from repo config. Example : 

```
hex, [{doc, ex_doc},
        {repos, [
                #{
                    name => <<"hexpm:myrepo">>,
                    api_url => <<"http://localhost:4000/api">>,
                    repo_url => <<"http://localhost:4000">>,
                    repo_verify => false,
                    repo_verify_origin => false
                },
                #{
                    name => <<"hexpm">>,
                    api_url => <<"http://localhost:4000/api">>,
                    repo_url => <<"http://localhost:4000">>,
                    repo_verify => false,
                    repo_verify_origin => false
                }]}]}.
```

paulo-ferraz-oliveira: Hm... so should it be a key in the map spec? Or why is the code working if I remove it (assuming it's tested, but probably it's not 😢)

starbelly: Well, it definitely works, I use it :) It is tested as part of the "integration" test suite as well. I suppose an interesting question, is why you expect it to fail? Perhaps you are thinking dialyzer should catch this?

paulo-ferraz-oliveira: > Well, it definitely works, I use it :)

I meant "I removed it, and it seems to 1. stop Dialyzer complaining, 2. not stop tests" (which is what I was referring to with the coverage part of my comment).

paulo-ferraz-oliveira: > why you expect it to fail?

If it's there (prior to this pull request) I'd expect it to fail if I removed it and it was required. Dialyzer complaining does not mean it's not working, it just means there's either a wrong spec or a wrong call...

paulo-ferraz-oliveira: > Perhaps you are thinking dialyzer should catch this?

Dialyzer definitely catches that we're calling a function with a map containing a key named `name` when it's not expected, but that doesn't mean there's an error, so I'm leaning toward an incomplete spec.

paulo-ferraz-oliveira: In summary: [in `hex_core`](https://github.com/hexpm/hex_core/blob/main/src/hex_core.erl#L90) the key appears to be named `repo_name` (not `name`), and it seems used [here](https://github.com/hexpm/hex_core/blob/main/src/hex_repo.erl#L197), but it's also possible that code is not used, and the name is always returned by the previous line of code.

paulo-ferraz-oliveira: Since this started to feel familiar, and went to search the code then the Git history, and I seem to have [made a similar change in the past](https://github.com/erlef/rebar3_hex/pull/229/files#diff-f91a12f48327c7c0b04d92bc558861bb13dbdbcff61f9f72f45c47ed2e8f8c1aR119), that got accepted and is in tag-released code. Should I create a function that identifies the goal of removing the key and use it in the existing places and the new places I just tweaked?

Also, is it possible we're using that key internally (`rebar3_hex` only) and not in `hex_core` since 1. there's working code without it, 2. Dialyzer complains, 3. tests don't fail without it... (?)

starbelly: Ahh I see, apologies, I should have been more clear. It is indeed internal to rebar3 and rebar3_hex, hex_core doesn't care about that key. Since the call would succeed, at the time it was deemed pragmatic vs trying to remove all keys that are not in `hex_core:config/0` for the sake of doing so.

paulo-ferraz-oliveira: Yeah, that's when I think by default Dialyzer should not care about extra keys in maps, but I can see benefits in both sides of the question.

paulo-ferraz-oliveira: `hex_core` has `repo_name`. Do you think it's retro-fittable, in that case? Stop using `name` and using `repo_name`?

src/rebar3_hex_client.erl
starbelly: Opened issue : https://github.com/hexpm/hex_core/issues/149

paulo-ferraz-oliveira: Opened pr 😄: https://github.com/hexpm/hex_core/pull/150

src/rebar3_hex_organization.erl
paulo-ferraz-oliveira: I've updated the current code (it seems `name` goes into `rebar3` which makes it more difficult to just use expected `repo_name`) by concentrating it in a single module: [src/rebar3_hex_client.erl](https://github.com/erlef/rebar3_hex/pull/338/files#diff-d27dd061e87ebcd5b13f06bad04d32afd9e3351ccbc771abab65d9e63de828c0).

starbelly: I did a bit of digging, if for nothing more than a stroll down memory lane :) Prior to 2019 there was no `repo_name`, anywhere, `name` was the cool kid on the block! :p  hex_core added `repo_name` in https://github.com/hexpm/hex_core/pull/48 . Likewise, this [https://github.com/erlang/rebar3/pull/1996] adjusted rebar to be compat with hex_core 0.4.0 and finally the first mention of repo_name in rebar3_hex is [here](https://github.com/erlef/rebar3_hex/pull/105). 

I think this was to disambiguate the parent (repository) from the child (org). 

And yes, I don't we can just stop using name. It's a worthy discussion to be had outside the scope of this PR though!

paulo-ferraz-oliveira: Got it. Lemme know if you prefer to ignore the Dialyzer, or keep the implementation I'm suggesting... (with the small wrapper function)

src/rebar3_hex_client.erl
paulo-ferraz-oliveira: I think now the "bug" boils down to what's more important (and that's probably interpreted differently by different people):

* is it the spec that's Ok and we should adapt the code (as per the pull request)
* is it the code that's Ok and we should adapt the spec (in which case it'll be less readable)

I personally think it doesn't hurt to support both the binary `rebar3_hex` is sending while making the public intention the atom and then manipulating the input inside the lib.

starbelly: I think that would be ok. Similar to what Eric stated on hex_core discussion, it is a bit more to maintain. By no means a heavy lift ever, but still, there is that. 

I suppose atm, we're waiting for your PR to get merged and new release of hex_core to be cut.

paulo-ferraz-oliveira: Yeah, I rebased over the main branch, and reversed this particular change (reason/message) so as soon as a new `hex_core` gets cut we can update here, then merge.

paulo-ferraz-oliveira: Resolve'ing, as this brings nothing new.

src/rebar3_hex_retire.erl
paulo-ferraz-oliveira: This is now outdated.

paulo-ferraz-oliveira:
CI's failing while we wait for an `hex_core` release. Then we update `rebar.config`+`rebar.lock` and retry 😄

paulo-ferraz-oliveira:
@starbelly, this is good to go. No more `-dialyzer` exceptions 😄

starbelly (approved):

PR #347, opened 2025-03-15
Don't tell user to run with DIAGNOSTIC=1 when running with DIAGNOSTIC=1

### LINKED ISSUE #344: Don't show DIAGNOSTIC message if run with it set
If there is an exception that isn't handled then even when running with `DIAGNOSTIC=` you get the message:

```
===> An unknown error was encountered. Run with DIAGNOSTIC=1 for more details.
```

I think this should be left out if its already enabled.

starbelly:
Closes #344

### DIFF
--- a/src/rebar3_hex_error.erl
+++ b/src/rebar3_hex_error.erl
@@ -32,4 +32,9 @@ format_error({Cmd, {error, {failed_connect, [{to_address,_},
 
 format_error(Reason) ->
     rebar_log:log(diagnostic, "Unknown error: ~p", [Reason]),
-    "An unknown error was encountered. Run with DIAGNOSTIC=1 for more details.".
+    case rebar_log:get_level() of
+        4 ->
+            "An unknown error was encountered";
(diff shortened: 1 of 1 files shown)


src/rebar3_hex_error.erl
```diff
@@ -32,4 +32,9 @@ format_error({Cmd, {error, {failed_connect, [{to_address,_},
 
 format_error(Reason) ->
     rebar_log:log(diagnostic, "Unknown error: ~p", [Reason]),
-    "An unknown error was encountered. Run with DIAGNOSTIC=1 for more details.".
+    case rebar_log:get_level() of
+        4 ->
+            "An unknown error was encountered";
+        _ ->
+            "An unknown error was encountered. Run with DIAGNOSTIC=1 for more details."
```
starbelly, line 39: It's arguable we should not do this at all. It was an attempt to be helpful, but perhaps it's a bit much.

ferd (approved):

PR #351, opened 2025-05-30
Show friendly error message when non-hex deps are found.

### LINKED ISSUE #343: Handle non_hex_deps exception
Trying to publish a package of an app with a non-hex dep rightly fails but the error isn't properly handled so you'll get the message to rerun with `DIAGNOSTIC=1` and then something like:

```
===> Unknown error: {create_package,
                        {error,{non_hex_deps,["eqwalizer_support"]}}}
```

starbelly:
Closes #343

### DIFF
--- a/src/rebar3_hex_build.erl
+++ b/src/rebar3_hex_build.erl
@@ -196,7 +196,7 @@ format_error({build_package, {error, Error}}) when is_list(Error) ->
 
 format_error({build_package,{error,{non_hex_deps,Deps}}}) ->
     DepsStr = lists:join(",", Deps),
-    Error = "Dependencies excluded from the package (only Hex packages can be dependencies) : ~s",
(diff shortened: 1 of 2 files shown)


src/rebar3_hex_build.erl
```diff
@@ -194,6 +194,11 @@ format_error({build_package, Error}) when is_list(Error) ->
 format_error({build_package, {error, Error}}) when is_list(Error) ->
     io_lib:format("Error building package : ~ts", [Error]);
 
+format_error({build_package,{error,{non_hex_deps,Deps}}}) ->
+    DepsStr = lists:join(",", Deps),
+    Error = "Dependencies excluded from the package (only Hex packages can be dependencies) : ~s",
```
paulo-ferraz-oliveira:
```suggestion
    Error = "Some dependencies were excluded from the package; only Hex packages can be dependencies: ~s",
```

paulo-ferraz-oliveira (approved):

PR #352, opened 2025-05-30
Update CI to reflect latest backwards compat support

starbelly:


paulo-ferraz-oliveira (approved):
`minimum_otp_vsn` still states `18`, but I'm not sure what the plan to move forward is... If we decide to move forward some code cleanup is in order 😄

starbelly:
> `minimum_otp_vsn` still states `18`, but I'm not sure what the plan to move forward is... If we decide to move forward some code cleanup is in order 😄

I need to revisit this. Specifically, if we stated that we would not support versions older than X going forward when rebar3_hex 7 was introduced.

PR #353, opened 2025-05-31
Raise error if no lock file detected when building.

### LINKED ISSUE #350: Pakage published to hex.pm with git dependency when no rebar.lock is present
I am still new to publishing packages with hex, this was my second publication, and the first with dependencies, so all of this may just be operator error.

I recently published an [update to `atomvm_rebar3_plugin`](https://hex.pm/packages/atomvm_rebar3_plugin/0.7.4) on hex. This release added a new dependency, which during development was only available via git. The dependency was already published to hex, but I forgot to change the source in the `atomvm_rebar3_plugin` rebar.config. We do not keep a rebar.lock file in our source repository, which turned out to be where my problem started.

I had been testing back and forth between OTP-27 and OTP-28 to ensure we were compatible with 28.0, but had not inadvertently broken older versions. In order to create a clean build I made sure to remove any un-tracked files and deleted my _build directory. This included the rebar.lock file. At this point running `rebar3 hex build` followed by `rebar3 hex publish --dry-run` did not indicate any errors, and the contents of my package looked correct when I examined it. I did not think to check metadata.config or the packaged rebar.lock. I proceeded to run `rebar3 hex publish`. This resulted in my package being published  without any dependencies listed for the release.

It did take me a while track down the fact that it is the lack of a rebar.lock, or a rebar.lock with just `[].` that would allow this to happen. I have since made sure to adjust my workflow to run `rebar3 compile` first to be sure the rebar.lock is updated before using `rebar3 hex build`, but it seems that the `hex build` task should either have a sanity check on the existence of a rebar.lock at the very least, or update rebar.lock like `rebar3 compile` does. I understand this might not be desirable behavior for some package maintainers, so a more robust solution might be to generate what hex thinks _should_ be in rebar.lock based on the deps, and raise an error if this does not match when a rebar.lock is present, otherwise generate the lock file.

Thank you for your time, I hope this is useful, and might help save someone else some head scratching.

starbelly:
Given that we can not check for the presence of non-hex deps without a lock file, we opt to bail if no lock file is detected on prior to creating a package, since under "normal" conditions the lock provider would have ran before the build and/or publish provider.

Closes #350

### DIFF
--- a/src/rebar3_hex_build.erl
+++ b/src/rebar3_hex_build.erl
@@ -211,8 +211,19 @@ format_error({build_docs, {error, missing_doc_index}}) ->
 format_error({build_docs, Error}) when is_list(Error) ->
     io_lib:format("Error building docs : ~ts", [Error]);
 
-format_error({create_package, no_lock_file}) ->
-  "No lockfile detected. Be sure to run rebar3 hex build package under the default profile prior to building or publishing under an alternative profile.";
+format_error({create_package, no_lock_file, default}) -> 
+  "No lockfile detected.";
+
+format_error({create_package, no_lock_file, CurrentProfiles}) ->
+    case lists:any(fun(P) -> P =:= lock end, CurrentProfiles) of 
+      true ->
+        "No lockfile detected.";
+      false -> 
+        Err = "No lockfile detected and one can not be generated running under the current profile.~n~n" 
+          ++  "Be sure to run rebar3 deps, rebar3 lock, or rebar3 hex build under the default ~n"
+          ++  "profile prior to building or publishing under an alternative profile.~n",
+          io_lib:format(Err, [])
+    end;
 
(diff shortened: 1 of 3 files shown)


src/rebar3_hex_build.erl
```diff
@@ -211,6 +211,9 @@ format_error({build_docs, {error, missing_doc_index}}) ->
 format_error({build_docs, Error}) when is_list(Error) ->
     io_lib:format("Error building docs : ~ts", [Error]);
 
+format_error({create_package, no_lock_file}) ->
+  "No lockfile detected. Be sure to run rebar3 hex build package under the default profile prior to building or publishing under an alternative profile.";
```
starbelly: This is wordy and a bit confusing.

ferd (approved):

PR #356, opened 2026-02-21
allow building with OTP-29

mikpe:
- fix old-style catch usage in `rebar3_hex_io` (it also had a broken match on `'Exit'` which should have been `'EXIT'`)
- enforce `nowarn_deprecated_catch` when building the `katana_code` dependency

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -8,6 +8,11 @@
     {platform_define, "^[2-9]", 'POST_OTP_18'}
 ]}.
 
+%% Some dependencies don't yet build cleanly with OTP-29.
+{overrides, [
+     {add, katana_code, [{erl_opts, [nowarn_deprecated_catch]}]}
+]}.
+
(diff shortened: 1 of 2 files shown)


rebar.config
```diff
@@ -8,6 +8,11 @@
     {platform_define, "^[2-9]", 'POST_OTP_18'}
 ]}.
 
+%% Some dependencies don't yet build cleanly with OTP-29.
+{overrides, [
+     {add, katana_code, [{erl_opts, [nowarn_deprecated_catch]}]}
```
tsloughter, line 13: This is a dependency of rebar3_hex?

mikpe: > This is a dependency of rebar3_hex?

Yes, via `rebar3_hank` and `rebar3_ex_doc`.

tsloughter: Aah, project plugins, missed those. Thanks!

tsloughter (approved):

